PR #13, opened 2013-06-11
Adding graphical tool to convert images to bitmap code

ehubin:
I have created a small java application that allows to quickly generate relevant char array to be used as a bitmap in Adafruit-GFX-Library.

Thought you might want to add it to the library files


Traceman:
Just needed to review.

Thank you!

PR #19, opened 2013-07-12
Please merge to add more flexibility

phibos:
Users can choose between 8-bit and 16-bit coordinate values and three different color modes.

Default modes should be compatible with all existing libs


PaintYourDragon:
Making a sweep through old PRs. Always grateful for contributions and the exchange of ideas, but won’t be merging this, would disrupt too many other libs, projects and user code, have to be careful as there’s a lot of unfixable stuff in print. Sometimes we “paint ourselves in a corner” with improvident design choices, and GFX being 16-bit-centric is one of those situations.

PR #32, opened 2014-04-17
Added support for multiple fonts, fast font rendering

oconnoda:
Changed font data from column major order to row major order.


PaintYourDragon:
Making a sweep through old PRs. Always grateful for contributions, but won’t be merging this since font support got merged into GFX somewhere along the way. Thank you.

PR #36, opened 2014-10-03
Draw lines as small line segments instead of pixels

pepijndevos:
It occurred to me that Bresenham's algorithm walks in the X and periodically steps the Y.
So I modified the algorithm to draw short lines instead of pixels.

The idea is that every time you increase the Y you store the current X, and when the Y is increased again, draw a line from the stored X to the current X.

The lines benchmark on my ILI9341 got about twice as fast, using my optimized fork adafruit/Adafruit_ILI9341#3

```
baseline
Screen fill              1135696
Text                     214436
Lines                    2009244
Horiz/Vert Lines         98852
Rectangles (outline)     68052
Rectangles (filled)      2358700
Circles (filled)         645056
Circles (outline)        877660
Triangles (outline)      637436
Triangles (filled)       1177460
Rounded rects (outline)  306064
Rounded rects (filled)   2665260

line segments
Screen fill              1135696
Text                     214436
Lines                    1065012
Horiz/Vert Lines         98852
Rectangles (outline)     68040
Rectangles (filled)      2358644
Circles (filled)         645060
Circles (outline)        877684
Triangles (outline)      242460
Triangles (filled)       1177452
Rounded rects (outline)  306068
Rounded rects (filled)   2665288
```

However, there is a off-by-one error somewhere. Some lines seem to wrap around the edge of the screen.
I've been staring at it for hours, so I'm just going to submit the pull.

img_20141003_160056

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -201,10 +201,11 @@ void Adafruit_GFX::drawLine(int16_t x0, int16_t y0,
       seg=x0+1;
     }
   }
+  // x0 incremented
   if (steep) {
-    drawFastVLine(y0, seg, (x0-seg)+1, color);
+    drawFastVLine(y0, seg, (x0-seg), color);
   } else {
-    drawFastHLine(seg, y0, (x0-seg)+1, color);
+    drawFastHLine(seg, y0, (x0-seg), color);
   }
 }
 


oliverab:
Maybe the base definition for hline, vline should be changed to a simple for-loop to prevent recursion. To me it makes more sense for a full line function to call a basic non-slope line function.

pepijndevos:
I fully agree. If there is any interest in merging this, I can look what is needed to make this work well, and merge cleanly.

oliverab:
I imported the library to MBED some time ago, and because I wasn't trying to maintain compatibility I was free to "enhance" it a bit. The main change I did was a coordinate swap so that it drew shapes using vlines not hlines. On the VFD I was using, GP9002A, that brings a massive speed enhancement as the byte-vertical memory organisation makes vlines crazy fast compared to hlines. That really needs to be switchable as some displays use byte-horizontal. Maybe a #ifdef switch?

The other things I changed were in the GP9002A code where I added a byte cache that greatly speeds up writing multiple pixels that are in the same byte.

I'm now looking at the possibility of converting it to C to run on an 8 bit PIC, I've seen other libs but they didn't include filled shapes.

PR #37, opened 2014-11-10
Made swap a function to prevent awful error messages from macros.

Chobbes:
This macro causes a lot of grief for poor students who aren't yet familiar with gcc's dreadful error messages. Unfortunately "swap" is a common name for functions, and if people aren't careful and declare their own functions when including Adafruit_GFX.h you get a misleading error, as opposed to a duplicate declaration one.

To clarify... I have seen this issue manifest itself many times in a university Arduino course! Preferably swap would not be exposed at all, but I worry that somebody might depend upon this.

### DIFF
--- a/Adafruit_GFX.h
+++ b/Adafruit_GFX.h
@@ -8,7 +8,7 @@
  #include "WProgram.h"
 #endif
 
-#define swap(a, b) { int16_t t = a; a = b; b = t; }
+inline void swap(int16_t &a, int16_t &b) { int16_t t = a; a = b; b = t; }
 
 class Adafruit_GFX : public Print {
 


fingolfin:
I think this is a good idea, though one should verify that the optimizer still manages to deal with this efficiently (it sure does on x86, but I have not tried to disasm the code on AVR?!?)

Chobbes:
Surely any modern C++ compiler should be able to optimize an inlined swap function :). I can't imagine there would be a performance regression, but I also don't have a dev environment set up to test this.

If somebody with merge privileges needs me to look at this, then I can. But it's been well over a year, so I'm not even sure if it would be looked at?

fingolfin:
I also agree that any modern C++ compiler *should* be able to optimize an inlined swap function. But I also learned the hard way that in embedded development, you should never rely soley on what you think the compiler *should* do, but should also verify it (it's pretty easy with `avr-objdump -S`).

That said, I am just a user (not a repo owner) of this project, and these are just my two cents. And yeah, it's not even clear whether they are interested in pull requests, given the lack of feedback on useful requests like this one :/.

Chobbes:
Yeah that's totally fair, I just don't have any avr stuff set up at the moment! (I would also be even less impressed with avr-g++ than I already am :P)

I'm not even sure where the appropriate place to put pull requests is either. There's a version of this library in the Arduino repositories as well, if I recall correctly.

PaintYourDragon:
Making a sweep through old PRs. Looks like the swap() macro got renamed somewhere along the way, so closing this. Always grateful for the extra eyes though, thank you.

PR #41, opened 2015-01-21
Custom Fonts

markslwong:
An interim solution until a true API is released.  Allows for alternate
fonts and font sizes  Fonts are restricted to being a fixed width for
all characters.  Use the Google GLCDFontCreator tool and strip all the
metadata before passing in the font.

Hint: You must modify the width of every character in the tool
Hint: The constant FONTNAME_WIDTH is not precise if you imported a font.
Hint: The first character specifies the first character to which the
data is supplied.  For example, by default the tool exports only
starting at character with ASCII code 32.

Note: I've kept the exact original behaviour.  There are some behaviours, (such as the reliance of the font data to have a blank space vertically), that should be fixed in separate commit.


ladyada:
heya if its ok - we added multifont support with fixed & variable-width support so will close this pull! thank you for your help :)

PR #42, opened 2015-02-01
Added support for program stored strings.

talbot-j:
Two functions added to support writing of strings stored in program memory.
The first function writes the string.  The second writes a string and a
line feed.

I find myself using these functions all the time to save RAM and thought others could take advantage of them as well.

Thanks for you consideration.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -419,7 +419,7 @@ void Adafruit_GFX::writePSTR( const char * c) {
 //Usage: display.writePSTRln( PSTR ( "Test" ) );
 //Notes: be sure to #include <avr/pgmspace.h>
 void Adafruit_GFX::writePSTRln( const char * c) {
-  writePSTR( * c );
+  writePSTR( c );
   write( '\n');
 }
 
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -403,6 +403,26 @@ void Adafruit_GFX::drawXBitmap(int16_t x, int16_t y,
   }
 }
 
+//Write character string from program memory to the display.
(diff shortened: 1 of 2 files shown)


fingolfin:
Printing of text is part of the `Print` class, see http://playground.arduino.cc/Code/Printclass resp. https://github.com/arduino/Arduino/blob/master/hardware/arduino/avr/cores/arduino/Print.h

So such a function should rather be added to the Print class, shouldn't it? But depending on your use case, it may not even be necessary, as you can already do
```
display.print(F("Some string"));
```
But this doesn't quite work if you have a progmem pointer which does not directly come from a string literal... Hmm. I guess then one could write
```
display.print(reinterpret_cast<const __FlashStringHelper *>(someVar));
```
but that's rather ugly... I am actually surprised that the `F` macro has no similar macro that just does this cast... Hmm

fingolfin:
Aha, see also https://github.com/arduino/Arduino/issues/2580

talbot-j:
If

`display.print(F("Some string"));`

prints "Some string" and "Some string" (with the NUL termination) is stored in the flash memory space (program memory) then this accomplishes the task intended by this pull request.  I am nearly 100% certain, outside of actually scanning the print.cpp source) that this F() feature was not documented when I submitted this pull.  I don't have time to test this, but if this works, and appears it will, then please cancel the pull and close it.  Thanks.

PaintYourDragon:
Making a sweep through old PRs. Always grateful for contributions, but decided against merging this. F() strings should cover the majority of cases, and the reinterpret_cast, though ugly, should cover the rest and essentially compiles to nothing rather than introducing redundant functions. Also, decreasingly relevant as things shift away from AVR. Thank you.

PR #56, opened 2015-08-23
graphic functions example

daniel3514:
Cheers!! Thanks for your great work
He added an example of the graphics library functions


PaintYourDragon:
Making a sweep through old PRs. Always grateful for contributions, but will not be merging this. Drawing examples accompany each hardware-specific library. Also, newlines have changed from original code resulting in unreadable massive diffs.

PR #68, opened 2016-01-02
Make custom fonts work on ESP8266 as well

markwal:
It looks to me like the ESP8266 has 32-bit pgm pointers


PaintYourDragon:
Much appreciated. Thanks!

PR #71, opened 2016-01-05
getTextBounds should consider the last line width

markwal:
In the default 5x7 font handling, the last line isn't considered when
determining the longest line.  Also, for the same reason, when there is
only one line, getTextBounds returns 65535 as the width.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -748,6 +748,7 @@ void Adafruit_GFX::getTextBounds(char *str, int16_t x, int16_t y,
     }
     // End of string
     if(lineWidth) y += textsize * 8; // Add height of last (or only) line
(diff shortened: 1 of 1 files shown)


skirchev:
X increment omitted in default fonts (should be incremented by the char width) and if the text does not end with new line the maxwidth should be replaced with linewidth.

markwal:
The code before my change uses 6 as the X increment.

The default font code (not mine, the code just above my change) sets lineWidth = to number of chars * 6.  It calculates the correct maxWidth if any wrapping occurs, it just leaves it at 0 if there was only one line worth of text.

markwal:
@PaintYourDragon @tdicola I think you want this fix for #81.  Feedback?

PaintYourDragon:
Merged, thank you!

PR #73, opened 2016-01-31
minor changes to build on Fedora

pljakobs:
fixed library includes, tested build on F21 and Cygwin


pljakobs:
just verified, this change is also necessary to build successfully on Linux Mint 17

PR #75, opened 2016-02-22
Added TomThumb (3x5 pixel font) to /Fonts

skelliam:
See comments in source.

The original Tom Thumb 3x5 font was (c)Brian Swetland, but modifications for readability are from (c)Robey Pointer.  Mr. Swetland used a BSD license (permissive) but Mr. Pointer would like to use a CC license, Mr. Swetland agreed.  See discussions below:
http://robey.lag.net/2010/01/23/tiny-monospace-font.html
http://opengameart.org/content/tom-thumb-tiny-ascii-font-3x5
http://opengameart.org/forumtopic/how-to-submit-art-using-the-3-clause-bsd-license

I've converted this font to use w/ Adafruit's GFX library, specifically targeting the 8x8 LED matrix display.  I've also decided to NOT make this a monospaced font; where possible the cursor will advance < 4 pixels  (i.e. the exclamation point, the number one, a comma, etc.).  

On the 8x8 matrix display this means you can quite easily display two numbers simultaneously.  But without monospacing, you can also display "111" if you wish.


PaintYourDragon:
Nice. Thanks!

PR #76, opened 2016-02-28
Added autoscrolling

resterampeberlin:
This pull request:

adds autoscrolling to the SSD1306 and GFX libraries. With this enhancement, a display can be used like a terminal because the displays scrolls automatically at the end of the display.

The function has been tested with the system and user defined fonts of different sizes and all orientations. I can provide test code if you want. 

The function works without interference to the rest of the library if the "autoscroll" is disabled.

Unfortunately the function requires, that scrolling is implemented in each subclassed class (like SSD1306). I could implement it but unfortunately I have only the 128x64 OLED display.


PaintYourDragon:
Making a sweep through some old PRs. Always grateful for contributions, but decided against merging this, since not all display hardware has physical scrolling capability. Forks are welcome and encouraged for special features for your own needs, upstream changes can then be kept in sync. Thank you.

PR #83, opened 2016-05-06
Modify type of SizeText in Adafruit_GFX.cpp 

marcbechi:
Change in function void Adafruit_GFX::setTextSize(uint8_t s) to Adafruit_GFX::setTextSize(float s). 
it's more convenient. 
For small screen if we want some text in medium size we don't have. 
if we choose the size 1 it's too small and size 2 it's too big so if we put a float instead of uint8_t we can choose something like 1.4 , 1.7 and it's more effective. 
Thank's

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -624,7 +624,7 @@ int16_t Adafruit_GFX::getCursorY(void) const {
   return cursor_y;
 }
 
-void Adafruit_GFX::setTextSize(uint8_t s) {
+void Adafruit_GFX::setTextSize(float s) {
   textsize = (s > 0) ? s : 1;
 }
 


WereCatf:
You do realize that this doesn't even work? All you do is change the parameter the function accepts, but none of the code that actually uses it! Besides, textsize is an uint8_t, so now the function will try to put a float into a uint8_t.

Then there's the whole thing that, well, most low-end MCUs that are used in Arduinos and the likes lack an FPU, so they'll have to use very slow, memory-hungry floating-point emulation, and so, if you went and modified the code to actually use floating-point maths, POOF -- you've just made the text-rendering of the library excruciatingly slow to the point of unusability.

PaintYourDragon:
For interim text sizes, please use the font support features added in v1.1 and documented here:
https://learn.adafruit.com/adafruit-gfx-graphics-library/using-fonts

As WereCatf points out, use of floating-point values in this microcontroller-centric code is discouraged for space and performance reasons.

PR #85, opened 2016-05-26
Make it compile on Non-Arduino platforms

cyraxx:
I'm working on an Embedded Linux system with a small SPI display and decided to use Adafruit GFX as a simple, light-weight drawing library. (To make it actually do something I will also develop and provide a subclass that writes to a Linux framebuffer.)

In order to make the library compile outside of the Arduino SDK I added a few tweaks to the code. Of course they only come into effect when `ARDUINO` is not defined.

If you'd like to keep your repository Arduino specific that's fine. I just figured a PR couldn't hurt since it's only minor tweaks and someone else might benefit from them at some point. But I can also keep them in my fork if you prefer.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -521,6 +521,13 @@ void Adafruit_GFX::write(uint8_t c) {
 #endif
 }
 
+#ifndef ARDUINO
+void Adafruit_GFX::print(const char *s) {
+  for(; *s != '\0'; s++)
+    write(*s);
+}
+#endif
+
 // Draw a character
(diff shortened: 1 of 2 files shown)


PaintYourDragon:
Making a sweep through some old PRs. Always grateful for contributions, but decided against merging this — not sure it’s needed anymore, plus might be more sensible to implement a minimal subset of the Print class that GFX inherits from, rather than putting these functions into GFX itself in special cases. Forks are welcome and encouraged for special features for your own needs, upstream changes can then be kept in sync. Thank you.

PR #86, opened 2016-06-01
Eliminated some sources of compiler warnings

jayeye:
Some variables were being used uninitialized, resulting in compiler warnings.

The first argument of getTextBounds() should be const, otherwise one cannot pass const objects to it.


PaintYourDragon:
Looks like these were resolved in an equivalent PR somewhere along the way. Thank you!

PR #95, opened 2016-10-30
Using fontconvert.c in Windows OS

johnbacas:
That's a brief guide of how to set up your windows OS PC to create new fonts with fontconvert.c

### DIFF
--- a/fontconvert/fontconvert_win.md
+++ b/fontconvert/fontconvert_win.md
@@ -45,7 +45,7 @@ make
 make install
 ```
 Once you're finished, go inside "C:\MinGW\include" and there should be a new folder named "freetype2".
-That propably means that you have installed the library correctly !!
+That, hopefully, means that you have installed the library correctly !!
 
 #### STEP 3: Build fontconvert.c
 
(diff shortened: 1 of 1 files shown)


microbuilder:
Looks good, thanks for taking the time to put this together.  Hopefully it will help other people in the same situation, and having a .md file in the repo means it can be updated over time if necessary.

pljakobs:
cool! glad to see you got it working!
Quick question: does the resulting binary need mingw to be installed or is it a windows standalone?
If it can live by itself, it might be a good idea to make it available as binary since not many windows users are privy with a development environment.

johnbacas:
Hi Peter
I've already made a pull request for this matter (#96).
The binary file (.exe) needs a .dll file to work with. In our case the file is "libfreetype-6.dll".
If "fontconvert.exe", "libfreetype-6.dll" and the .ttf file are in the same folder then someone can use the command prompt to run "fontconvert.exe" with the necessary parameters and create the header file.
I've tested on a pc without MinGW and it worked !!

PR #97, opened 2016-10-31
Improved speed of drawChar using custom fonts by using fillRect

jaretburkett:
Switched drawChar with custom fonts to use fillRect on a per line basis. This increases the speed on avr slightly, but gives an enormous speed increase when using a library that uses DMA in the fillRect function.


PaintYourDragon:
Making a sweep through some old PRs. Looks like equivalent functionality was covered somewhere along the way. Thank you!

PR #99, opened 2016-11-09
Too early text wrap fix

egzumer:
`cursor_x + textsize * 6` is last pixel to be written, it can be equal `_width`

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -480,7 +480,7 @@ void Adafruit_GFX::write(uint8_t c) {
     } else if(c == '\r') {
       // skip em
     } else {
-      if(wrap && ((cursor_x + textsize * 6) >= _width)) { // Heading off edge?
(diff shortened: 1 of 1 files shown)


maxint-rd:
Tested again just now in version 1.1.9 of Adafruit GFX library. Not fixed yet.
Please consider changing line 483 of Adafruit_GFX.cpp to:
      ``` if(wrap && ((cursor_x + textsize * 6) > _width)) { // Heading off edge? ```
(so all what is needed is to change the >= to >)

maxint-rd:
Tested version 1.2.2. Now I think it's fixed (see https://github.com/adafruit/Adafruit-GFX-Library/commit/d0e78f9db12a72d5ead2fb19f392ccfa0fd5379e#diff-3f1cb26c6b84c9aa9fe99816de4f3fb9)

PaintYourDragon:
Making a sweep through some old PRs. Indeed, looks like an equivalent fix got incorporated somewhere along the way. Thank you!

PR #101, opened 2016-12-09
Update Adafruit_GFX.cpp

u48:
Fix Adafruit_GFX::write()
It was not a valid comparison of the character position and the right edge of the display.


u48:
fix printing last char in line

PR #102, opened 2016-12-10
Width and Height of a button should be 16 bit, not 8 bit

ghost:
The display has a height/width of 320 Pixel.
But the button height and width can only be a uint8_t.
Internally the width and height are already 16 bit.

So the init button function should also support 16 bit values


PaintYourDragon:
Making a sweep through some old PRs. Appears that an equivalent change was made in another PR along the way. Much appreciated, thank you!

PR #105, opened 2017-02-01
Fix matc next simbol position in function write()

u48:
Fix matc next simbol position in function write().
Otherwise, the last character is not drawn in a line (for which sufficient space).


u48:
If this idea is promising, then I can write variants for ARM,ESP8266.

PaintYourDragon:
Making a sweep through some old PRs. Looks like an equivalent fix was made in another PR along the way. Much appreciated, thank you!

PR #107, opened 2017-02-28
Implement transaction based drawing

me-no-dev:
With using transactions to draw, we save huge amount of time in pin
changing and SPI setup. This new model, together with the updated
ILI9341 lib give a much better performance.


Adafruit_GFX.cpp
```diff
@@ -71,596 +71,680 @@ POSSIBILITY OF SUCH DAMAGE.
 #endif
 
 Adafruit_GFX::Adafruit_GFX(int16_t w, int16_t h):
-  WIDTH(w), HEIGHT(h)
```
Bendito999: The diff here is strange with spacing differences from the previous one along with many other spots, is this intentional?

me-no-dev: not really, it's more of auto-spacing of the editor. There were many mixed spaces/tabs and that is not nice to look at and read. This diff is best to be looked at as result rather than diff. Some methods are moved and some are added. Git diff is confusing sometimes.

Bendito999: Ah, okay, as long as it is moving in the direction of more consistent than less consistent, great!

PR #108, opened 2017-03-02
Add missing const get functions for TextColor, TextSize, and TextWrap.

jrcutler:
These tend to be needed in generic code that needs to restore previous state.


PaintYourDragon:
Making a sweep through some old PRs and I apologize that this got overlooked. Text sizing has since changed (now X & Y scales) and it looks like a later PR covers this & will likely merge. No shade intended, grateful for the contribution. Thank you!

PR #114, opened 2017-04-09
gettextbounds fix

stanzlavos:
When using default font, wrong width was being returned.


PaintYourDragon:
Making a sweep through some old PRs. Looks like this got addressed somewhere along the way with other changes in getTextBounds(). Much appreciated, thank you.

PR #115, opened 2017-04-10
Added getTextSize

help137:
for use in routines such as calculating text placement.


PaintYourDragon:
Making a sweep through some old PRs. Text scaling has since changed (separate X & Y) and a later PR provides compatible getter calls, so I’ll be merging that. No shade intended, appreciate the contribution, thank you.

PR #119, opened 2017-04-22
fixed strict-aliasing warning

grafalex82:
pgm_read_pointer() calls in Adafruit_GFX.cpp generate number of warnings like this:

warning: dereferencing type-punned pointer will break strict-aliasing rules
#define pgm_read_dword(addr) (*(const unsigned long *)(addr))

I noticed this when building the library for STM32 microcontroller with -Wall switch

In this pull request I did 2 things:
1) moved pointer magic with pgm_read_* calls into functions. For AVR microcontrollers code should be identical. I hope compiler is smart enough to inline these functions (at least ARM gcc is)
2) added a special case for non-AVR controllers that does not use pgm_* calls at all


grafalex82:
I refreshed this PR according to changes in master. CI is green.
Please consider taking this change to master.

makermelissa:
Hi, thanks for doing this. Which displays have you tested this with?

grafalex82:
I tested with SSD1306 and SSD1309 displays, using Arduino Nano as well as STM32F103 (via STM32GENERIC) microcontroller

makermelissa:
Ok thanks. There's a few more displays I'd like to test with before merging. :)

PR #124, opened 2017-05-17
Create Tiny3x3a2pt7b

eadmaster:
Smallest tiny font i could find.
Allows showing 3+ lines of text on a 32x16 matrix (not very readable but still nice to have).


PaintYourDragon:
I was gonna say 'no' because that's just ridiculous...but I CAN kinda read it, it's like an accent. :) And thank you for setting a good example with the attribution!

PR #125, opened 2017-05-20
Bitmaps XOR compression implemented (-38,73% in average 1st invention, *** improved up to -50% for tested fonts)

eltomjan:
Bitmaps of 48 fonts now consume 110073B / original 179672 B.
Bits of rows are XORed, there is a lot of zeros then and they are replaced by 0 bit in data (remaining non-zeros data use 9 bits).
Data are placed in a new container (cArrayPtr.h) during encoding, fixed 8 bits are followed by non-zero bytes (zeros are skipped), so best option is 0x00 for 8 zeros in a row and worst 0xFF and 8 non-zero bytes.
Encoder is there:
https://github.com/eltomjan/ETEhomeTools/blob/master/gfxXorPacker.cpp
Developed and tested on my ArduinoSimulator 1st and then in Pololu A-Star 32U4 Micro & Adafruit_ILI9340 today.

BR,
El Tom Elzein
P.S. Bitmap ptrs can be stored relative to each other and point first to 8b bytes map and then to particular byte index (0-7) - in case 64k/8 = 8192 would not be less than last char index as it looks like sometimes.


romansavrulin:
@eltomjan Good work! Can you update existing encoder to convert directly from ttf?

eltomjan:
Just published C# translation (https://github.com/eltomjan/ETEhomeTools/blob/master/fontReader.cs, https://github.com/eltomjan/ETEhomeTools/tree/master/CSharp), it is a bit refactored, so quite simple to understand...
It should be much easier to integrate it somewhere else...

eltomjan:
Finally done... Made it as a refactoring - original code sends data to binary stream now instead of direct print and final output is redirected to file.
It was next to impossible to refactor origina "spaghetti code"...
Results seems to be same as compressing of existing font, but did not test result on Arduino not simulator. There is also a "demo" text print of "raw" font content in new comment (string txt) / line 530.

eltomjan:
Removed older ugly web commits and made 2 new versions using Git (merge with gfxXorPacker.cpp / XOR & encoding layer in original fontconvert.c).

eltomjan:
Improved version - there is 1B with block size & compressed bitstream (0 => block size of zero bits / 1 => block size unencoded bits - ones or mixed). Made merge with latest branch, but there are some bugs somwhere else...

eltomjan:
Now even integration machine seems 2 B broken.
Latest version is encapsulated to separate objects, so minimal changed logic inside drawChar.
Still decoding original uncompressed chacters/format also.
Compression even aprox. 10% better thanx 2 variable block (instead former 8b) size and uncompressed store when compression does not help.

eltomjan:
Hi @PaintYourDragon, can this PR be merged? And if not, can you explain why so I can fix or rewrite it?

PaintYourDragon:
The roadmap (in README.md) explains that one of the primary goals going forward is not to break users' existing code or data. As some users have already converted a number of fonts for their needs, this change would break their data. Also mentioned there is a desire NOT to bloat out the code with additional formats for fonts or bitmaps (frankly wish I could REMOVE some bitmap formats, but...would break some users projects)...so...what's in there might not be perfect, but is adequate for most needs and continues to work without breaking compatibility.

If you require a more nutrient-dense font format for your projects, I'd recommend making a fork of the library...you take can the font stuff wherever you need for your projects, won't break others' projects, and can still track other changes in the mainline code to stay up-to-date with the remainder of the library.

eltomjan:
@PaintYourDragon sorry 4 misunderstanding, but let me ask once more - if you find excellent font compression method exceeding even LZW - will you throw it out after your project or would you look for a way to give it to others, who also fight 4 each B in flash sometimes ? I am new in open source, so do not know how / where to publish that... And whole arduino is full of bloat - even simple string steal few kB w/o reason and out of control - would guess you can simply delete say 2/3 of code/size you do not use in your project if there will be complete src translated each time. Btw also write pixel is very slow - do not need to sent useless coords with each pixel thru slow SPI - significant speedup is to send square of all bits at once for bigger fonts.

eltomjan:
Btw, in case there will be a flag or 0. bitmap offset/size will be used to distinguish, it can work with both formats or also very simple to make conversion - already have simple source file reader in C# for (import's fonts from source file for R&D purposes https://github.com/eltomjan/ETEhomeTools/blob/master/CSharp/Program.cs).

pljakobs:
Tom, I'm trying to build a more feature rich fork of the gfx library. At this point, it doesn't have (m)any takers, but maybe if there were more contributors, it might become a valuable "unstable" development fork.

pj

eltomjan:
Thanx, I am open for improvements or adjustments of my compression code and if you prefer to work on PC (windows VS Express C++ .Net for example), I have also Arduino simulator here https://github.com/eltomjan/ETEhomeTools/tree/master/ArduinoSimulator but as it is on different platform it should not be a big problem to siwtch Qt, Mono, etc.

pljakobs:
so I have tried to implement a 24Bit color model and bitmapped canvasses here https://github.com/pljakobs/Adafruit-GFX-Library/tree/Feature-24Bit
As said, I'd be most happy to have this become a "what Arduino based graphics could be" branch.

pj

eltomjan:
Now have also simple GFX editor in JScript (legacy format + convertion of glyphs table option) here https://github.com/eltomjan/ETEhomeTools/blob/master/HTM_HTA/GFX/fontShowConvert.htm

PaintYourDragon:
Making a sweep through some old PRs. I appreciate the hard work that’s gone into this, but one of the points specifically mentioned in the roadmap is that we don’t wish to break compatibility with existing fonts in projects, nor add more formats. If your projects require the compressed format, forks are welcome and encouraged, and any upstream changes can then be kept in sync. Thank you.

PR #126, opened 2017-05-23
Set yAdvance to glyph height if face height is zero

rbultman:
Bitmapped PCF and BDF font files might not (don't?) have face height information.  If the face height is zero from the face metrics, use the height of a glyph as the yAdvance.  This keeps yAdvance from being zero, which allows line-wrapping to work correctly.

Also added .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+default.vim
+fontconvert/fontconvert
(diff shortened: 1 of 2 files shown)


PaintYourDragon:
I did not know that! Thanks for the fix.

PR #127, opened 2017-06-30
create clearDisplay() and display() functions

pljakobs:
I'm working on an abstract gauges library that will be able to show information on any sort of display. It uses Adafruit_GFX as the abstraction to access the display itself. 
However, currently, the interfaces differ because some driver classes decide to impelment additional functions that are not defined within Adafruit_GFX itself. Since I don't necessarily know if I'm drawing on an 1306 oled or a ILI9340 tft and can't find out on the code level, I think implementing virtual dummy functions in Adafruit_GFX is the right thing to do.
I have defined 
Adafruit_GFX::display() as doing nothing as it's not needed functionality for most displays and overwritten for those who need it
Adafruit_GFX::clearDisplay() calls fillScreen(0) as there's no concept of a screen background color (only a font background color).


pljakobs:
ping. any chance to include this very simple PR? it would allow people to use my Gauges library without having to use a special AdafruitGFX patch

Issue #90, opened 2016-08-23
getTextBounds should have String as param as well


xmenxwk:
How to resolve this ?

```

    String someText = "SomeText";//generated by a method
    int16_t x1, y1;
    uint16_t w1, h1; 
    tft.getTextBounds(someText , 0, 0, &x1, &y1, &w1, &h1);//error here
```

Tried using c_str();, doesn't help either.

amitkumarsharma008:
int str_len = someText .length() + 1; 
    char char_array[str_len];
    someText .toCharArray(char_array, str_len);
    lcd.getTextBounds(char_array, 1, 56, &upperleftX, &upperleftY, &width, &height);

minida28:
Usage of `String` is considered as a bad practice in embedded devices with limited memory, so it should be avoided as hard as possible. Maybe @ladyada will not implement your request due to that reason.

Here is a good reading:

[https://hackingmajenkoblog.wordpress.com/2016/02/04/the-evils-of-arduino-strings/](https://hackingmajenkoblog.wordpress.com/2016/02/04/the-evils-of-arduino-strings/)

ladyada:
hiya - yes we strongly prefer no Strings plz :)

vortigont:
I'm not promoting Strings here. Actually using this lib it is possible to .print() a String to the LCD,
but at the same time you can't find it's size with .getTextBounds(). This doesn't look consistent at all when some methods works on the object and some doesn't. If we do not like Strings in one place than why use it in others? Besides, the fix that I offered doesn't use Strings manipulation at all, it just casts it as a const char pointer, enough to calculate WxH on the screen.

ladyada:
ooh right. its been a while since we wrote that code. @vortigont can you rebase since we've made a ton of changes and it isnt a clean merge?

dragondaud:
I manually patched to use the String overload, and would like to see that merged. The argument against Strings is fine on tiny microcontrollers, but the $2 ESP8266 has plenty of RAM, so the point is moot.

vortigont:
@ladyada  done. pls, review.
thanks for looking into this.

ladyada:
https://github.com/adafruit/Adafruit-GFX-Library/pull/130#issuecomment-412334177

PR #130, opened 2017-08-29
Added String overload to getTextBounds() #90

vortigont:
While it is possible to .print() a String, inability to do .getTextBounds() over String is very inconvenient.
consider this
```
String mytext = "SomeText";
lcd.print(mytext);       // this one works
lcd.getTextBounds(mytext, 0, 0, &x1, &y1, &w1, &h1);    //this one does not
```

So we have to use a temp buffer every time and copy String to char*, like this
```
char char_array[mytext.length() + 1];
mytext.toCharArray(char_array, mytext.length());
lcd.getTextBounds(char_array, 0, 0, &x1, &y1, &w1, &h1);
```

This simple patch adds String overload to getTextBounds() without using a temp buffer. Should  be backward compatible with older code. (addressing Issue #90)

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1415,7 +1415,18 @@ void Adafruit_GFX::getTextBounds(const char *str, int16_t x, int16_t y,
     }
 }
 
-// Overload for the same as above, but for String strings
+/**************************************************************************/
+/*!
+    @brief    Helper to determine size of a string with current font/size. Pass string and a cursor position, returns UL corner and W,H.
+    @param    str    The ascii string to measure (as an arduino String() class)
+    @param    x      The current cursor X
+    @param    y      The current cursor Y
+    @param    x1     The boundary X coordinate, set by function
+    @param    y1     The boundary Y coordinate, set by function
+    @param    w      The boundary width, set by function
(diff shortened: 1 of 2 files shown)


ladyada:
hiya needs some documentation and we can merge  once travis passes :)
https://travis-ci.org/adafruit/Adafruit-GFX-Library/builds/412688931?utm_source=github_status&utm_medium=notification

vortigont:
@ladyada how can I help with documentation? not sure what should I do?

ladyada:
you need to doxygen the functions - you can look at other functions to see the formatting, we also have some text here https://learn.adafruit.com/the-well-automated-arduino-library/doxygen

vortigont:
OK, got it! Thanks!

vortigont:
@ladyada docs added. Not sure how doxygen differentiate docs for overload functions, I've just copy'ed the one for the same function and added a note about String() class as an argument.

ladyada:
lovely!

vortigont:
Thanks @ladyada. I think this merge covers issues #90, #149, #159, perhaps it might be closed now.

ladyada:
wow thanks vorti! best contributer ever :)

PR #132, opened 2017-09-02
Autoscroll support. No conflicts.

shprota:
This is just a cleanup for the #76 
It should be applied together with [Adafruit_SSD1306#56](https://github.com/adafruit/Adafruit_SSD1306/pull/56)
I think, many people would be glad to have this feature finally available.


shprota:
Thanks for pointing out to the wrapping bug. Fixed that in PR.

PR #135, opened 2017-09-27
Add support for ESP-IDF

me-no-dev:
This PR adds component.mk so the library can be used as a component inside ESP-IDF+Arduino projects


andrzejtokarski:
Too bad it is still not merged!

me-no-dev:
I guess after so much time, we can now also add CMake config to the waiting list as well :D

ladyada:
hihi i missed this one - you could add a makefile...but it seems like every library would need this. can you check what platformio does? because i know they solved this sort of problem and without needing extra files!

ladyada:
ok i will await ... but.... we wont be maintaining any extra files - that is, if we update a library, we wont be testing/checking whatever you add, so please make sure you're ready to update them yourself or make them loosely defined enough they dont need maintanence

me-no-dev:
I understand :) then I think the best option is to close the request. Otherwise any change in the file structure/names will cause errors. Nobody wants that. I will be looking for another way.

ladyada:
yeah - i want to avoid future pain since we have done some pretty big refactors and we only have travis CI for arduino regression. thank you! :)

PR #141, opened 2017-11-20
增加五角星和椭圆函数

ytshang123:


PaintYourDragon:
Making a sweep through some old PRs. Always grateful for contributions, but this is sufficiently esoteric that it’s best handled in user code. Thank you.

PR #150, opened 2018-02-16
Compiler warnings fixed (signed / unsigned mismatch)

VioletGiraffe:
A trivial edit that fixes the only two compiler warnings emitted by this library.


PaintYourDragon:
Making a sweep through some old PRs. Looks like this got addressed by another PR along the way. Thank you!

PR #152, opened 2018-03-06
A proposal for forcing custom fonts to a font grid - to allow drawing…

bishibashiB:
… background without artifacts
This apporach accepts cutting oversized letters (bigger than x/yAdvance) for the sake of having a background color, no flickering on updating (due to first blockout and redraw).
The feature can be enabled by calling _setCustomFontToGrid_()  and therefore backward compatible.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -80,6 +80,7 @@ WIDTH(w), HEIGHT(h)
     textsize  = 1;
     textcolor = textbgcolor = 0xFFFF;
     wrap      = true;
+	forceCustomFontToGrid = false;
     _cp437    = false;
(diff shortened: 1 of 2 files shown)


Jekyll555:
Used the adoptions within this project 
[https://github.com/Jekyll555/Twizy_CanDisplay](Twizy_CanDisplay )

PaintYourDragon:
Making a sweep through old PRs. Grateful for the contribution, but we’ve settled on the Official Canonical Way To Handle Non-Blinking Overwrites With Custom Fonts is to use an interim GFXcanvas1 object. If grid-spaced text is required, can use one of the 'FreeMono' fonts, or convert another monospaced font. Thanks.

Issue #149, opened 2018-02-07
Why does getTextBounds() not take a const char*?


Sod-Almighty:
I cannot pass a `const char*` to `getTextBounds()`. Why not?

pljakobs:
'''
getTextBounds(char *string, int16_t x, int16_t y,
      int16_t *x1, int16_t *y1, uint16_t *w, uint16_t *h),
    getTextBounds(const __FlashStringHelper *s, int16_t x, int16_t y,
      int16_t *x1, int16_t *y1, uint16_t *w, uint16_t *h);
'''
It's only defined for char* and flash strings. If you define a const char*, it's probably best to define it as flash string anyway, that way it'll reside in progmem which is larger on most platforms. 

pj

Sod-Almighty:
Perhaps I should be more specific. I already _knew_ the function signatures; what I _meant_ was "why didn't you make the function accept a `const char*` instead, given that it doesn't modify the string?"

I'm using an ESP8266. Not only does it have plenty of RAM, it also doesn't seem to properly support the `Flash` library or the `F()` macro.

Sod-Almighty:
I have never heard of such a limitation. But in any event, nothing stops there being two versions of the function, one const and one non-const. Having to explicitly cast away constness is _bad programming practice_.

pljakobs:
@Sod-Almighty I'm sure you are aware of the fact that this is open source software being developed in a collaborative way. You are of course free to for it, write the suitable overloads for your needs, test them and create a PR.

vortigont:
I've done this PR #130 quite some time ago, it also allows to pass String() to getTextBounds()

ladyada:
https://github.com/adafruit/Adafruit-GFX-Library/pull/130#issuecomment-412334177

PR #159, opened 2018-04-22
Change getTextBounds() from char * to const char * to avoid warnings/errors when passing a const char to it.

leres:
I like making char * arguments to functions that do not modify the argument const. But if I have a routine that calls getTextBounds() with a const char *, for example:
```
static int
mono_stringwidth(const char *s)
{
        int16_t x, y;
        uint16_t w, h;

        display.getTextBounds(s, 0, 0, &x, &y, &w, &h);
        return (w);
}
```
I get this error:
> invalid conversion from 'const char*' to 'char*' [-fpermissive]

Changing the char * version of getTextBounds() to be const char * solves this issue and works when passing a char * too. Not only will you get an error if you try to modify the const char * by mistake, it can also helps the compiler because it knows this argument cannot be changed.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -996,7 +996,7 @@ void Adafruit_GFX::charBounds(char c, int16_t *x, int16_t *y,
 }
 
 // Pass string and a cursor position, returns UL corner and W,H.
-void Adafruit_GFX::getTextBounds(char *str, int16_t x, int16_t y,
+void Adafruit_GFX::getTextBounds(const char *str, int16_t x, int16_t y,
         int16_t *x1, int16_t *y1, uint16_t *w, uint16_t *h) {
     uint8_t c; // Current character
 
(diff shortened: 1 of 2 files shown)


vortigont:
I've made a PR #130 some time ago addressing the same issue (it also allows passing String() to getTextBounds()), but it's still not merged.
Also referencing this issues #149 #90

pljakobs:
I think it would be good to have an "unstable" branch/fork for AdafruitGFX that is a little more liberal in accepting PRs than what @PaintYourDragon has laid out under "roadmap" in the readme.
There are a number of good PRs that have not made it into mainline, and I have a number of (major) things lined up as well.
While we probably all have our forks, I certainly don't have the time to fully maintain a testing fork, but if maybe two or three people got together, I think we could drive this forward.
Thoughts?

kind regards

pj

pljakobs:
I think it would be good to have an "unstable" branch/fork for AdafruitGFX that is a little more liberal in accepting PRs than what @PaintYourDragon has laid out under "roadmap" in the readme.
There are a number of good PRs that have not made it into mainline, and I have a number of (major) things lined up as well.
While we probably all have our forks, I certainly don't have the time to fully maintain a testing fork, but if maybe two or three people got together, I think we could drive this forward.
Thoughts?

kind regards

pj

ladyada:
https://github.com/adafruit/Adafruit-GFX-Library/pull/130#issuecomment-412334177

PR #169, opened 2018-07-27
esp use_fast_pinio undef

tablatronix:
This was never an issue until include macro was moved to below the definition, masked before 1.2.6
via

https://github.com/adafruit/Adafruit-GFX-Library/commit/b2cc77a13b8b35689a99a4c0a7eb9c5d186f8edc#diff-34d217f1a3e35d00f24b2c3a7bca0dad

### DIFF
--- a/Adafruit_SPITFT.h
+++ b/Adafruit_SPITFT.h
@@ -22,6 +22,7 @@ typedef volatile uint32 RwReg;
 typedef volatile uint32_t RwReg;
 #elif defined(ESP32) || defined(ESP8266)
 typedef volatile uint32_t RwReg;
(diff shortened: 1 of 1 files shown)


tablatronix:
USE_FAST_PINIO does not work for esp, but was enabled in 1.2.6

WereCatf:
Why disable fast pinio on the ESP, instead of fixing it?

ladyada:
we never used it, it also doesnt work for some pins!

WereCatf:
I haven't checked on the ESP32, but they all certainly work just fine on the 8266.

tablatronix:
Care to elaborate?

PR #171, opened 2018-07-29
Adafruit_GFX::write implementation should match declaration for older Arduino platform

nayanshah:
Adafruit_GFX.h has conditional declaration of write method which returns a value only on newer Arduino. 
Change definition of write in Adafruit_GFX.cpp to be void for older platforms.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1118,7 +1118,11 @@ void Adafruit_GFX::drawChar(int16_t x, int16_t y, unsigned char c,
     @param  c  The 8-bit ascii character to write
 */
 /**************************************************************************/
+#if ARDUINO >= 100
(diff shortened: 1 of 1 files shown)


ladyada:
hiya we dont support arduino < 1.0, there's been enough changes and years gone by :)

nayanshah:
@ladyada : For context, we're trying to create an "emulator" for the displays (e.g. Adafruit_SharpMem). Intent is to be able to test changes rapidly while developing without needing to flash to a microcontroller. 

An early version is available here: https://github.com/nayanshah/Adafruit-SharpMem-Emulator

It builds Adafruit_GFX by including the repo as a submodule and forwards drawPixel to a canvas. I don't care about older Arduino versions, but just using that check to include custom logic.

PR #178, opened 2018-09-13
Add the ability to pass in a SPIClass (Sercom) to use for HW SPI.

daniel-stewart:
This change allows a SPIClass pointer to be passed in and used as the HW SPI. Allowing a SPIClass to be given opens the door to using SPIs (and associated Sercoms) that are not the default SPI as defined in varaints.cpp. With this change, SPIs that are added in an Arduino sketch and not defined in variant.cpp can be used. It was this change (and associated change in the ST77xx drivers) that allowed me to use the UncannyEyes with the Circuit Playground.


ladyada:
done!

Issue #180, opened 2018-09-23
please separate Adafruit_SPITFT from Adafruit_GFX library


ZinggJM:
I used Adafruit_GFX with many different Arduino IDE platforms.
It used to compile for all platforms I used without problems.
But now it no longer compiles for some platforms because of the added Adafruit_SPITFT part.
Platforms affected I encountered so far are for STM32: STM32DUINO and STM32GENERIC.

Adafruit_GFX is the base graphics library used for Adafruit libraries, but also for many other libraries.
Adafruit_SPITFT is a useful addition for some libraries, but should be in its own library.

Thank you for your great work!

ZinggJM:
#define RwReg uint32_t
in 
C:\Users\xxx\Documents\Arduino\hardware\STM32GENERIC\STM32\cores\arduino/stm32/stm32_gpio.h:43:15:
conflicts with 
  typedef volatile uint32_t RwReg;
in 
C:\Users\xxx\Documents\Arduino\libraries\Adafruit_GFX_Library\Adafruit_SPITFT.h:23:29:
and
   #include "wiring_private.h"
does not exist for some platform

ladyada:
i think this is a pretty minor change, and i'd like to keep SPITFT in here because 99% of GFX users will also use SPITFT - could you submit a PR that fixes this for STM32? we dont own any STM32 boards to test with :)

WereCatf:
I cannot find any reference to stm32_gpio.h in the Arduino OTTO - core that's installable straight from the Arduino IDE by default, in Roger Clark's STM32Duino - core or in the ST-official STM32Duino - core, so I can't do anything about that, but fixing the issue with "wiring_private.h" is simple:

In Adafruit_SPITFT.cpp change the line `#ifndef RASPI` to `#if !defined(RASPI) && !defined(STM32F4)` and the compile-error goes away in all the cores I just tested.

ZinggJM:
stm32_gpio.h is in https://github.com/danieleff/STM32GENERIC
https://github.com/danieleff/STM32GENERIC/blob/master/STM32/cores/arduino/stm32/stm32_gpio.h
and RwReg is still defined as a macro in the actual version, unfortunately.
So I would propose to replace RwReg by RwReg_t in Adafruit_SPITFT (as it is a typedef).

WereCatf:
Did a pull-request; compiles fine, didn't test on actual hardware. Leaving that part to @ZinggJM

[...]

PR #181, opened 2018-09-26
Attempt to fix compile on STM32GENERIC #180

WereCatf:
Briefly tested compilation on STM32GENERIC, Roger Clark's Stm32duino, ST official Stm32duino and Arduino Star OTTO. Did not test on actual hardware.


PaintYourDragon:
Looks like this got resolved through other changes somewhere along the way. Pretty sure the standard GFX lib will compile in this situation now. Closing issue, please re-open if that’s not the case. Thank you.

Issue #88, opened 2016-06-21
ft2build.h  file is missing


HP-Sparks:
Hi,
I'm giving a try to a simple Arduino-based Oscilloscope project, the " fontconvert.c " file refers to a " ft2build.h " file to be included. (see line 23: `#include <ft2build.h>`)

Where can we find this file if Not present amongst the already present libraries on the computer ?

Thanks in advance.

Best regards,
  HP_

pljakobs:
Hi HP_,

what platform are you building on?
you'll need the freetype 2 development packages under Linux / cygwin.
I have tested it on Fedora 21, SuSE 13.x, Mint and cygwin.

pj

HP-Sparks:
Hi @pljakobs ,
Sorry, my bad... I was so focus on my work and the parallel hunts for data, files and information that I forgot to share this information... :-/

Well, I'm using the Arduino IDE 1.6.9 on a Linux Ubuntu-based distribution system.

After installing the latest stable IDE from the official Web site [(http://arduino.cc),](http://arduino.cc) I allowed all the automatic updates requests I faced for the bundled and newly available libraries.

Then, I also did a search on both Internet and my system, it seems that this file disappeared for many people for some reason(s). (copyright/licensing issue ?? trying to guess...)

Best regards,
HP_

HP-Sparks:
Hi @pljakobs,

Sorry, my bad... I was so focus on my work and the parallel hunts for data, files and information that I forgot to share this information... :-/
Well, I'm using the Arduino IDE 1.6.9 on a Linux Ubuntu-based distribution system.

Now, going to search for _Font Type 2_ development packages...
But, yesterday, searching on my system (using `me@netbook / # find / ft2build.h
`) or on Internet, I found nothing except many people facing the same issue for all kind of projects developed using all kind of programming languages...
So, I'm wondering if this file has been removed due to copyright/licensing issue... ?? (trying to guess...)

Thank you for the clue ! :-)

Best regards,
HP_

**P.S.**
Just in case there would be a question on the Arduino IDE:
 - After installing the latest stable IDE from the official Web site [(http://arduino.cc),](http://arduino.cc) I allowed all the automatic updates requests I faced for the bundled and newly available libraries.
 - No magic from this too. ;-)

HP-Sparks:
Well, I found something from there: http://download.savannah.gnu.org/releases/freetype/freetype-2.6.tar.gz (latest version at the moment)
[...]

PR #183, opened 2018-10-17
Fixed error from fontconvert.c when compiling example sketch of adafruit OLED display

theintel:
**Issue:**
When compiling the [example sketch](https://github.com/adafruit/Adafruit_SSD1306/blob/master/examples/ssd1306_128x64_spi/ssd1306_128x64_spi.ino) for adafruit OLED display, the sketch refers to Adafruit_GFX and consequently to fontconvert.c
This file includes a header file named ft2build.h, which many of the compilers fail to find, resulting in compilation error.

Check the discussion on this issue: #88

**Fix:**
Simply adding two lines at the beginning and at the end solves the problem: https://github.com/adafruit/Adafruit-GFX-Library/commit/4a2757c8934461a6936cec47a9c05980031441d8

### DIFF
--- a/fontconvert/fontconvert.c
+++ b/fontconvert/fontconvert.c
@@ -16,6 +16,7 @@ Keep 7-bit fonts around as an option in that case, more compact.
 
 See notes at end for glyph nomenclature & other tidbits.
 */
+#ifndef ARDUINO
 
 #include <stdio.h>
 #include <ctype.h>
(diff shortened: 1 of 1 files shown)


ladyada:
thanx!

theintel:
> thanx!

Heya! Thanks for being so quick in approving the pull request. This is why Iike Adafruit.

ladyada:
thanks - when its simple we can merge fast :)

PR #185, opened 2018-11-14
add UTF-8 fonts compatibility

mayabushlyakova:
This small fix will allow to use UTF-8 fonts. 

Maybe i will add some of my utf-8 fonts later...

### DIFF
--- a/gfxfont.h
+++ b/gfxfont.h
@@ -21,8 +21,8 @@ typedef struct {
 typedef struct { 
 	uint8_t  *bitmap;      ///< Glyph bitmaps, concatenated
 	GFXglyph *glyph;       ///< Glyph array
-	uint8_t   first;       ///< ASCII extents (first char)
-        uint8_t   last;        ///< ASCII extents (last char)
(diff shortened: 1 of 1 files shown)


pljakobs:
UTF-8 fonts don't lend themselves easily to small memory applications as many of them have non-contiguous glyph tables.
Nevertheless, i18n is important for me, too, but I would suggest that character sets following iso-8859-x would be the better approach as that would keep it to 255 characters per font.

Tijgerd:
maby you can add an:  #if defined(SUPPORT_UTF8)

idea--list:
I do not think iso-8859-x would be a good idea. Simply because that is an even bigger pitfall than good old ASCII. I mean if you can just use ASCII you know what you may expect. Choosing ISO now would just pave the way to hell and make porting to UTF-8 at some point even harder.

I agree with @Tijgerd and UTF-8 should be an opt-in turned off by default but commented properly.

PaintYourDragon:
Thank you for the contribution. Will be closing this PR and studying #200 which incorporates a similar change and takes it a step further.

PR #189, opened 2018-11-28
change gfxfont.h

AnatoliyKube:
Hello
I want to connect the font 12 * 7, as the main one, using my pseudographic characters.
What needs to be done so that the font is displayed correctly?
Changes to: Adafruit_GFX.cpp
void Adafruit_GFX::drawChar


PaintYourDragon:
Don't do this in the library. #include the font header in your Arduino sketch and call setFont().

PR #192, opened 2018-12-10
Cleanup: Remove `boolean` type usage and bizarre function forward definition.

fake-name:
Currently, this library uses the arduino `boolean` type, which was (and is) basically a really terrible idea that shouldn't have ever happened (how was that ever a thing?). The PR gets rid of it, because c++ has a native `bool` type.

Also, the function definitions in the header were just, well, *weird*, using one return type for a giant list of functions. This PR also cleans that up, so that you have properly independent definitions for each function.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -766,7 +766,7 @@ void Adafruit_GFX::drawBitmap(int16_t x, int16_t y,
 
 /**************************************************************************/
 /*!
-   @brief      Draw PROGMEM-resident XBitMap Files (*.xbm), exported from GIMP. 
+   @brief      Draw PROGMEM-resident XBitMap Files (*.xbm), exported from GIMP.
    Usage: Export from GIMP to *.xbm, rename *.xbm to *.c and open in editor.
    C Array can be directly used with this function.
    There is no RAM-resident version of this function; if generating bitmaps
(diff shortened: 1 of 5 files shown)


PaintYourDragon:
bool vs boolean: Arduino is full of bizarro decisions like this. This is an Arduino-specific library, and as such, being a “good neighbor” means following along with the conventions of their universe, for better or for worse. As it is, boolean true and false map to 1 and 0…but who knows, maybe next week they’ll decide it needs to be an enumeration. Sticking with their convention would make such a change less problematic should something like that occur.

“Bizarre” function declarations: this is entirely valid C syntax. Functions, like variables (as a special case of variable, really), can be declared in comma-delimited lists.

fake-name:
Even the *arduino* people say [you shouldn't use `boolean`](https://www.arduino.cc/reference/en/language/variables/data-types/boolean/)

> boolean is a non-standard type alias for bool defined by Arduino. It’s recommended to instead use the standard type bool, which is identical.

------

> “Bizarre” function declarations: this is entirely valid C syntax. Functions, like variables (as a special case of variable, really), can be declared in comma-delimited lists.

Just because it compiles doesn't make it a good idea. You could use goto everywhere too. The point is it's a unusual, idosyncratic way of doing things, and is the opposite of being a good neighbor.

PaintYourDragon:
Good to know boolean is officially deprecated now; will keep that in mind going forward.
The function declarations are fine.

fake-name:
Ok, W/E. I Tried. If you don't want to improve gross code, that's your prerogative.

PR #196, opened 2019-01-12
added function scrollUp() to scroll display with println()

psionprime:
This is my second pull request to a public project, so hopefully it's fine, otherwise let me know so I can improve. I tried implementing this keeping in mind other products but I am not sure this will be okay. By all means, have a more experienced person change what is required.

This was developed on a Adafruit Huzzah32 Feather stack with an OLED 128x32 Wing using the Adafruit Arduino support libraries on Win10x64 MS Visual Studio 2017 Community Edition with VisualGDB extensions supporting my Segger J-Link for JTAG debugging.

This function is called from Adafruit_GFX (pull coming soon)::write() to scroll the display if cursorY is greater than the display height. It will also call the driver display() function. This function is to allow the print system to display as many other terminals do. Printing without \n will work as expected. I can see adding a flag to toggle this behavior.

- only supporting standard font for ssd1306 OLED 128x32 as that is all I have to test on

Thanks Adafruit for the openness of your systems !

Attached is an Arduino test sketch. Please remember you need my fork of Adafruit_SSD1306 for it to work fully otherwise, with the empty function nothing much will happen:
[scroll-print.zip](https://github.com/adafruit/Adafruit_SSD1306/files/2752608/scroll-print.zip)

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -221,10 +221,18 @@ void Adafruit_GFX::endWrite() {
 
 /**************************************************************************/
 /*!
-   @brief    Scroll the display up one textsize line
+   @brief    Scroll the display up one textsize line; virtual to be implemented in driver
 */
 /**************************************************************************/
 void Adafruit_GFX::scrollUp() {
+}
+
+/**************************************************************************/
+/*!
+   @brief    update display; virtual to be implemented in driver
+*/
+/**************************************************************************/
+void Adafruit_GFX::display() {
 }
 
 /**************************************************************************/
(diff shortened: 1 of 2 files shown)


PaintYourDragon:
Making a sweep through some old PRs today. Very grateful for the work and consideration here, but I’ve closed out a couple of other scrolling-related PRs, and will this one as well, as adding this feature at this point is inherently problematic — either requiring scrolling support in the hardware (not universally handled), framebuffer-readback capability (ditto) and/or the addition of virtual functions to every single GFX subclass, and there’s quite a few now. Like some other painted-into-corner limitations of GFX, prob best to accept it’s just not happening, at least in the main repo. Folks are always welcome and encouraged to create their own forks though, say if the hardware scope of a project makes scrolling practical, and upstream changes can be kept in sync pretty easily then. Thank you.

PR #199, opened 2019-02-09
Add capability for using extended character sets

Bodmer:
UTF-8 decoder added to print stream so the UTF-8 encoded Unicode strings produced by the compiler are rendered correctly on the graphical displays.

Example for ILI9341 TFT showing printing Hiragana characters to screen here:

[hiragana_example.zip](https://github.com/adafruit/Adafruit-GFX-Library/files/2848193/hiragana_example.zip)

This update makes glyphs from the entire 16 bit Unicode Basic Multilingual Plane available. See comments in the example. This example uses ~45% of FLASH on an UNO so it should run fine on any hardware. The included font files were generated by fontconvert tool with the Unicode Hiragana code point range of 12353 - 12435.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1124,11 +1124,76 @@ void Adafruit_GFX::drawChar(int16_t x, int16_t y, unsigned char c,
 }
 /**************************************************************************/
 /*!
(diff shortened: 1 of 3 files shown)


Bodmer:
Ooops, this is an early version and has bugs.

PR #200, opened 2019-02-09
Add capability to use custom fonts that include characters from the Unicode Basic Multilingual Plane

Bodmer:
UTF-8 decoder added to print stream so the UTF-8 encoded Unicode strings produced by the compiler are rendered correctly on the graphical displays.

Example for ILI9341 TFT showing printing Hiragana characters to screen here:

[hiragana_example.zip](https://github.com/adafruit/Adafruit-GFX-Library/files/2848269/hiragana_example.zip)

This update makes glyphs from the entire 16 bit Unicode Basic Multilingual Plane available. See comments in the example. This example uses ~45% of FLASH on an UNO so it should run fine on any hardware. The included font files were generated by fontconvert tool with the Unicode Hiragana code point range of 12353 - 12435.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1124,11 +1124,76 @@ void Adafruit_GFX::drawChar(int16_t x, int16_t y, unsigned char c,
 }
 /**************************************************************************/
 /*!
+    @brief  Serial UTF-8 decoder
+    @param  c  8 bit value from encoded stream
+    @returns   0 if decoding is not complete yet, 16 bit code point
+               otherwise. Can cast to 8 bits for ASCII range (0-255)
+*/
+/**************************************************************************/
+
+uint16_t Adafruit_GFX::decodeUTF8(uint8_t c)
+{
+    // 7 bit Unicode Code Point
+    if ((c & 0x80) == 0x00) {
+        decoderState = 0;
+        return (uint16_t)c;
+    }
+
+    if (decoderState == 0)
+    {
+        // 11 bit Unicode Code Point
+        if ((c & 0xE0) == 0xC0)
+        {
+            decoderBuffer = ((c & 0x1F)<<6); // Save first 5 bits
+            decoderState = 1;
+            return 0;
+        }
+
+        // 16 bit Unicode Code Point
+        if ((c & 0xF0) == 0xE0)
+        {
+            decoderBuffer = ((c & 0x0F)<<12);  // Save first 4 bits
+            decoderState = 2;
+            return 0;
+        }
+
+        // 21 bit Unicode  Code Point not supported so fall-back to extended ASCII
+        if ((c & 0xF8) == 0xF0) return (uint16_t)c;
+    }
+    else
+    {
+        if (decoderState == 2)
+        {
+            decoderBuffer |= ((c & 0x3F)<<6); // Add next 6 bits of 16 bit code point
+            decoderState--;
+            return 0;
+        }
+        else // decoderState must be == 1
+        {
+            decoderBuffer |= (c & 0x3F); // Add last 6 bits of code point
+            decoderState = 0;
+            return decoderBuffer;
+        }
+    }
+
+    decoderState = 0;
+
+    return (uint16_t)c; // fall-back to extended ASCII
+}
+
+
+/**************************************************************************/
+/*!
     @brief  Print one byte/character of data, used to support print()
     @param  c  The 8-bit ascii character to write
 */
 /**************************************************************************/
 size_t Adafruit_GFX::write(uint8_t c) {
+  
+    c = (uint8_t)decodeUTF8(c);
+    
+    if ( c==0 ) return 1;
+
     if(!gfxFont) { // 'Classic' built-in font
 
         if(c == '\n') {                        // Newline?
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1018,13 +1018,13 @@ void Adafruit_GFX::drawRGBBitmap(int16_t x, int16_t y,
    @brief   Draw a single character
     @param    x   Bottom left corner x coordinate
     @param    y   Bottom left corner y coordinate
-    @param    c   The 8-bit font-indexed character (likely ascii)
+    @param    c   The 16-bit font-indexed character
     @param    color 16-bit 5-6-5 Color to draw chraracter with
     @param    bg 16-bit 5-6-5 Color to fill background with (if same as color, no background)
     @param    size  Font magnification level, 1 is 'original' size
 */
 /**************************************************************************/
-void Adafruit_GFX::drawChar(int16_t x, int16_t y, unsigned char c,
+void Adafruit_GFX::drawChar(int16_t x, int16_t y, uint16_t c,
   uint16_t color, uint16_t bg, uint8_t size) {
 
     if(!gfxFont) { // 'Classic' built-in font
@@ -1066,7 +1066,7 @@ void Adafruit_GFX::drawChar(int16_t x, int16_t y, unsigned char c,
         // newlines, returns, non-printable characters, etc.  Calling
         // drawChar() directly with 'bad' characters of font may cause mayhem!
 
-        c -= (uint8_t)pgm_read_byte(&gfxFont->first);
+        c -= pgm_read_word(&gfxFont->first);
         GFXglyph *glyph  = &(((GFXglyph *)pgm_read_pointer(&gfxFont->glyph))[c]);
         uint8_t  *bitmap = (uint8_t *)pgm_read_pointer(&gfxFont->bitmap);
 
@@ -1185,12 +1185,12 @@ uint16_t Adafruit_GFX::decodeUTF8(uint8_t c)
 /**************************************************************************/
 /*!
     @brief  Print one byte/character of data, used to support print()
-    @param  c  The 8-bit ascii character to write
+    @param  utf8  The 8-bit UTF-8 or ASCII code
 */
 /**************************************************************************/
-size_t Adafruit_GFX::write(uint8_t c) {
+size_t Adafruit_GFX::write(uint8_t utf8) {
   
-    c = (uint8_t)decodeUTF8(c);
+    uint16_t c = decodeUTF8(utf8);
     
     if ( c==0 ) return 1;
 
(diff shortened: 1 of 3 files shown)


Bodmer:
Screenshot from test sketch:
hiragana_test

Bodmer:
I suspect that to get around the lack of UTF-8 decoding in the library folk are having to use a work-around and bypass the print class by calling the write() member function directly.

So, the current pull request needs to have a UTF-8 on/off switch included.  The cp437() on/off switch function does not have a a generic name so, unfortunately, a further function will need to be added somehow.  Perhaps a gfxConfig() function is the best approach as that can also be used for future switchable extensions too, perhaps associated with graphics, stream and character set handling options.

pljakobs:
Hi Bodmer, I'm glad to see someone taking on the extended character set topic.
I had been thinking about this for years but never gotten around to actually doing it.
One of my concerns was the required flash space for a full 16 bit slice so I thought it would be more apropriate to implement iso8859 codepages.
I will pull this into my experimental fork and have a look at it.

Bodmer:
OK, I have added a switch function to my local copy so will push that to Git in the next few hours.

My thoughts are:
1. To make the minimum changes needed and (importantly) to make complete use of existing library features, **but** not preclude functional expansion at a future time (e.g.21 bit code point)
2. No definitive numerical point code boundary constraints or size limts on the glyph block start and end (no power of 2 constraints on boundaries). **But** start with a 16 bit Unicode code point variable.
3. Allow any number of separate contiguous point code blocks, initially from the Basic Multilingual Plane, to be loaded as arrays in FLASH
4. Make the library user repsonible for restricting the range/size and number of glyph blocks according the hardware they are using and it's capabilities.
5. Support UTF-8  (aka use encoded Unicode) on the print stream only, via the write() function
6. Add a UTF-8 on/off switch (default is UTF-8 off for legacy sketch support) to alow direct use of 8 bit write calls to access the top 128-255 characters of the GLCD legacy font.
7. Keep the UTF-8 decoder simple and not try to manage corruptted data streams
8. Make the drawChar function volatile, so users can create extension libraries that for example to: add diacritical marks to characters, write characters right to left or up to down, add their own font formats, to use alternative storage media for the fonts etc etc. Thie drawChar will be fed decoded Unicode code points.

This means the update can be transparent to existing users until deliberately switched on but that the rich character/glypg/grapheme set from the Basic Multilingual plane is available if wanted.  The Arduino IDE/GCC compiler already uses UTF-8 encoding for the print stream so this is the obvious way forward to me. The GFX "free fonts" will thus be still availble in the basic 32-127 range and in blocks anywhare in the 18 bit code point space.  The code size impact is minimal.

Bodmer:
I have uploaded the latest changes.

The copy of the[ ILI9341 library here](https://github.com/Bodmer/Adafruit_ILI9341) has two new examples that demo the added features. Adapting for other display hardware should be straightforward.

idea--list:
@Bodmer That is simply awesome!

Would pretty much produce my own font.h file. Installed MinGW, newest Freetype, etc. Can generate fonts (fontconvert myfont.ttf 12 32 512 > myfont), however characters beyond 0xFF simply do not render if using my own fonts. While your hiragana fonts work ok.
If i check my own fonts with [this tool](https://tchapi.github.io/Adafruit-GFX-Font-Customiser/), beyond 0xff i find hex ids that consist of only 3 digits like 0x100 etc. While UTF-8 fonts should have a four digit hex identifier if i am right.

So my question is how can i generate fonts.h that has special character identifiers in the proper hex format?

Bodmer:
The font data has been created following the instructions here:
  [https://www.youtube.com/watch?v=L8MmTISmwZ8](https://www.youtube.com/watch?v=L8MmTISmwZ8)

Note that online converters for Adafruit_GFX compatible fonts are
available but these typically only use characters in the range 32-127,
and thus do not include the accented characters. These online converters
can however still be used with this sketch but the example characters
used must be changed.

idea--list:
First thanks for the quick reply!

Yes i saw that video yesterday and right now i also found [this link](https://github.com/Bodmer/Adafruit-GFX-Library) of yours.
There you say you used this command to generate the hiragana.h font:
fontconvert C:\Windows\Fonts\sourceFont.ttf 12 12353 12435 > hiragana12.h

just did the same with my arial.ttf font and now even my glyphs have hex ids like 0x3041.
I can confirm it by checking it [here](https://tchapi.github.io/Adafruit-GFX-Font-Customiser/). of course in arial.ttf there was no such character but i can redraw it in that tool and save the modified arial.h file. Now by pasting 0x3041 hex id in the arduino sketch the display renders exactly the glyph i just drawed.

However if i build the font with this command:
fontconvert C:\Windows\Fonts\arial.ttf 12 32 512 > arial.h
I get a file that contains a lot of real glyphs (as arial.ttf is a UTF8 font) however this time most characters have a hex id like 0x123. If i paste such hex ids in the arduino sketch then something weird gets displayed despite the font.h file contains the glyph.

For me it seems as a possible bug in GFX library, i mean it seems to be able to lookup 1 byte characters that have a hex id between 0x00 and 0xff. It can also look up 2 byte characters with hex ids between 0x0000 and 0xffff (like in your hiragana font). But it seems to find glyphs correctly if they have hex ids between 0x000 and 0xfff.

While converting font.ttf files shipped with windows, after 0xff (=255 dec) the next glyph id is always 0x100 (=256 dec) but those do not get rendered right while the font.h contains most special fonts exactly in that range...

Can you please confirm this on your side or tell me what i might be missing?

Bodmer:
Create an example complete with font file for me to test. Post it here by making a zip file and dragging to the Git Write window.

idea--list:
Here it is:

[UTF8_test.zip](https://github.com/adafruit/Adafruit-GFX-Library/files/3537650/UTF8_test.zip)

Bodmer:
There are quite a few problems with your sketch.

1. A char array should be zero terminated otherwise you will likely get garbage characters at the end of the string.  So add a zero thus:
  char testch_ff[] = {'A','b', 0x21, 0x26, 0x4d, 0x7e, 0xff, 0x00};

2. A char array can only hold 8 bit values, so if you try to load it with a larger value then only the least significant byte will be stored. So you line thus:
  char testch_ffff[] = {0x3041, 0x3042, 0x3043, 0x3044, 0x3045};
Will actually create an array:
  char testch_ffff[] = {0x41, 0x42, 0x43, 0x44, 0x45};

3. To use UTF8 encoded characters you must switch on UTF8 decoding with a line thus:
   display.setAttribute(UTF8_ENABLE , true);
  To use a font with only characters in the range 0-0xff turn it off:
  display.setAttribute(UTF8_ENABLE , false);

4. To use 16 bit Unicode characters you have to use either the character itself in a string, e.g "こんにちは" or you must use the universal character name format e.g.:
 display.println("\u3041\u3042\u3043\u3044\u3045");
 display.println("Works: \u0100\u0106\u0124\u015c\u00ff");           // ALWAYS define as 16 bit hex

Make these changes to your sketch and see if it works.

idea--list:
@Bodmer i could not stop and i think i got it working now ;-)

[Here ](https://github.com/Bodmer/Adafruit-GFX-Library)you have a modified version of v1.3.6 of the GFX library, but that does not compile on my board.

So now i updated v1.5.6 with your modifications (will open a PR in your branch) and now it compiles to my board and i can get any custom characters displayed if i pass them as char array or as String.

Passing by hex code still does not work however, but i think i can live without that :-)

idea--list:
@Bodmer Thanks for taking your time and for the hints!

idea--list:
@Bodmer just created PR #244 where i describe everything and mention your work and help.
Thanks again!

MOSHEMEL:
HI bodmer  I built fonts as you said using the fontconvert  app for hebrew letters and  I checked it using the Adafruit GFX Pixel font customiser and I can see the font from 0x5d0 -0x5ea but my problem is when using tft.print("\u05D0\u05D1"); I do not see different letter but the same letter  
Do I need to take your  Adafruit_gfx  libray?

MOSHEMEL:
or  I need to do something else?

Bodmer:
You would need to download the [adapted library here](https://github.com/Bodmer/Adafruit-GFX-Library).

Note that the adapted library is a temporary proof of concpt library and that Adafruit have not adopted the changes needed in their master library and I will not be able to resolve any issues you have with it.

I would use [this library](https://github.com/olikraus/U8g2_for_Adafruit_GFX#u8g2_for_adafruit_gfx) to extend the fonts and permit use of UTF8 encoded strings. That library is well written and is designed to be used with the original Adafruit_GFX library.

Jarvis785:
Hi @idea--list and @BillyDonahue .
Please help me out, i'm trying out Devanagari UTF font in neomatrix, it doesn't work ;( 
I've downloaded you pull request.
`// Adafruit_NeoMatrix example for single NeoPixel Shield.
// Scrolls 'Howdy' across the matrix in a portrait (vertical) orientation.

#include <Adafruit_GFX.h>
#include <Adafruit_NeoMatrix.h>
#include <Adafruit_NeoPixel.h>

#include <Fonts/gargi.h>
#define FONT_NAME Amiko_Regular9pt8b //Devanagari font Range 0900–097F

#ifndef PSTR
 #define PSTR // Make Arduino Due happy
#endif

#define PIN 8

Adafruit_NeoMatrix matrix = Adafruit_NeoMatrix(12, 14, PIN,
  NEO_MATRIX_BOTTOM     + NEO_MATRIX_LEFT +
  NEO_MATRIX_ROWS + NEO_MATRIX_ZIGZAG,
  NEO_RGB            + NEO_KHZ800);

const uint16_t colors[] = {
  matrix.Color(255, 0, 0), matrix.Color(0, 255, 0), matrix.Color(0, 0, 255) };

void setup() {
  matrix.begin();
 // matrix.setFont(&FONT_NAME);
  //matrix.setTextWrap(false);
  matrix.setBrightness(70);
  matrix.setTextColor(colors[0]);

 matrix.fillScreen(matrix.Color(255,140, 0));// Test scree by filling colour(Working)
  matrix.show();
  delay(1000);
}

int x    = matrix.width();
int pass = 0;

void loop() {
  matrix.fillScreen(0);
  matrix.setCursor(x, 7);
  matrix.println(("अंतरिक्ष यान से दूर नीचे पृथ्वी शानदार ढंग से जगमगा रही थी ।"));//Not working
 //matrix.println(F("ABC"));//This is working perfecty fine
 
  if(--x < -36) {
    x = matrix.width();
    if(++pass >= 3)
    pass = 0;
    matrix.setTextColor(colors[pass]);
  }
  matrix.show();
  delay(100);
}`

colesnicov:
can I write Cyrillic?

DoomHammer:
@Bodmer can you resolve the merge conflicts?

Bodmer:
@DoomHammer 
I think that would be nugatory effort given that this pull request is 3 years old!

DoomHammer:
I believe this year to be the year of UTF-8 on microchips.

colesnicov:
> I believe this year to be the year of UTF-8 on microchips.

I've used mcufont.

PR #209, opened 2019-04-13
added the sendCommand function

makermelissa:
I created a non-const and const version as we discussed and also an overloaded one that allows just the command byte for functions like the SSD1351 invert, which I tested.

### DIFF
--- a/Adafruit_SPITFT.cpp
+++ b/Adafruit_SPITFT.cpp
@@ -1640,7 +1640,7 @@ uint16_t Adafruit_SPITFT::color565(uint8_t red, uint8_t green, uint8_t blue) {
  */
 void Adafruit_SPITFT::sendCommand(uint8_t commandByte, uint8_t *dataBytes, uint8_t numDataBytes) {
     SPI_BEGIN_TRANSACTION();
-    SPI_CS_LOW();
+    if(_cs > 0) SPI_CS_LOW();
   
     SPI_DC_LOW(); // Command mode
     spiWrite(commandByte); // Send the command byte
@@ -1651,7 +1651,7 @@ void Adafruit_SPITFT::sendCommand(uint8_t commandByte, uint8_t *dataBytes, uint8
       dataBytes++;
     }
   
-    SPI_CS_HIGH();
+    if(_cs > 0) SPI_CS_HIGH();
     SPI_END_TRANSACTION();
 }
 
@@ -1663,7 +1663,7 @@ void Adafruit_SPITFT::sendCommand(uint8_t commandByte, uint8_t *dataBytes, uint8
  */
 void Adafruit_SPITFT::sendCommand(uint8_t commandByte, const uint8_t *dataBytes, uint8_t numDataBytes) {
     SPI_BEGIN_TRANSACTION();
-    SPI_CS_LOW();
+    if(_cs > 0) SPI_CS_LOW();
   
     SPI_DC_LOW(); // Command mode
     spiWrite(commandByte); // Send the command byte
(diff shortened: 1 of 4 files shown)


makermelissa:
I'm not sure what's going on with Travis for the Zero, but it compiled in travis on my repo and also compiles fine on a computer.

makermelissa:
Ok, looks like it had a bad Zero Board package cached. Temporarily commenting out in .travis.yml, committing, then uncommenting and committing seems to have forced it to fix itself.

ladyada:
you can delete the cache in travis - please do that instead of changing travis.yml :)

makermelissa:
Oh sweet. I didn't realize that. Thanks!

Adafruit_SPITFT.cpp
```diff
@@ -1632,6 +1632,51 @@ uint16_t Adafruit_SPITFT::color565(uint8_t red, uint8_t green, uint8_t blue) {
...
 
+/*!
+ @brief   Adafruit_SPITFT Send Command handles complete sending of commands and data
+ @param   commandByte       The Command Byte
+ @param   dataBytes         A pointer to the Data bytes to send
+ @param   numDataBytes      The number of bytes we should send
+ */
+void Adafruit_SPITFT::sendCommand(uint8_t commandByte, uint8_t *dataBytes, uint8_t numDataBytes) {
+    SPI_BEGIN_TRANSACTION();
+    if(_cs >= 0) SPI_CS_LOW();
```
ladyada: change to _cs > 0, as pin #0 is a valid pin

Adafruit_SPITFT.cpp
```diff
@@ -1632,6 +1632,51 @@ uint16_t Adafruit_SPITFT::color565(uint8_t red, uint8_t green, uint8_t blue) {
...
+    SPI_DC_LOW(); // Command mode
+    spiWrite(commandByte); // Send the command byte
+  
+    SPI_DC_HIGH();
+    for (int i=0; i<numDataBytes; i++) {
+      spiWrite(*dataBytes); // Send the data bytes
+      dataBytes++;
+    }
+  
+    if(_cs >= 0) SPI_CS_HIGH();
```
ladyada: ditto

Adafruit_SPITFT.cpp
```diff
@@ -1632,6 +1632,51 @@ uint16_t Adafruit_SPITFT::color565(uint8_t red, uint8_t green, uint8_t blue) {
...
+
+/*!
+ @brief   Adafruit_SPITFT Send Command handles complete sending of commands and const data
+ @param   commandByte       The Command Byte
+ @param   dataBytes         A pointer to the Data bytes to send
+ @param   numDataBytes      The number of bytes we should send
+ */
+void Adafruit_SPITFT::sendCommand(uint8_t commandByte, const uint8_t *dataBytes, uint8_t numDataBytes) {
+    SPI_BEGIN_TRANSACTION();
+    if(_cs >= 0) SPI_CS_LOW();
```
ladyada: ditto

Adafruit_SPITFT.cpp
```diff
@@ -1632,6 +1632,51 @@ uint16_t Adafruit_SPITFT::color565(uint8_t red, uint8_t green, uint8_t blue) {
...
+  
+    SPI_DC_LOW(); // Command mode
+    spiWrite(commandByte); // Send the command byte
+  
+    SPI_DC_HIGH();
+    for (int i=0; i<numDataBytes; i++) {
+      spiWrite(pgm_read_byte(dataBytes++)); // Send the data bytes
+    }
+  
+    if(_cs >= 0) SPI_CS_HIGH();
```
ladyada: ditto

makermelissa: 👍

Adafruit_SPITFT.cpp
makermelissa: 👍

Adafruit_SPITFT.cpp
makermelissa: 👍

Adafruit_SPITFT.cpp
makermelissa: 👍

PR #212, opened 2019-04-13
add clearDisplay() and display() member functions. Also adds const datatypes to getTextBounds

pljakobs:
I hope this is now better than the first one. For some reasons, I was not able to just add the commits for clearDisplay() and display() but I guess the const string types won't break a thing and they have been requested before

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -283,6 +283,15 @@ void Adafruit_GFX::fillScreen(uint16_t color) {
     fillRect(0, 0, _width, _height, color);
 }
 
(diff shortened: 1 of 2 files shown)


pljakobs:
@ladyada I *believe* the diff is correct, rebasing my fork was a bit of an issue. I think I should learn more git :-o
I'm not sure if the failing checks are due to this diff, that more looks like general issues.

ladyada:
gettextbounds stuff is also being included but we really want to stick to just one change per PR since tens of thousands of people use this code. can you make a new branch, add just the code for display/fillscreen, and PR from that?

PR #217, opened 2019-05-06
New Figure

Soyunr:
ydpgsj


ladyada:
elipse is good, but pentagram is a bit specialized, want to turn it into drawNgram?

Soyunr:
drawPentagram()&drawEllipse()

Issue #218, opened 2019-05-09
fillTriangle() fails on 320x480 displays


prenticedavid:
There is intermediate overflow in the "lower half" of a triangle.
It can be fixed by casting the calculation to 32-bit:

```
    // For lower part of triangle, find scanline crossings for segments
    // 0-2 and 1-2.  This loop is skipped if y1=y2.
    sa = (int32_t)dx12 * (y - y1);
    sb = (int32_t)dx02 * (y - y0);

```
David.

ladyada:
woops - wanna try submitting a pr and @makermelissa can test? :)

makermelissa:
Fixed by #219

PR #219, opened 2019-05-09
Fix overflow in fillTriangle()

ImpulseAdventure:
Addresses integer overflow fix identified by @prenticedavid in #218


makermelissa:
I’ll test this when I get home.

PR #226, opened 2019-06-12
Text magnification factor can be defined separately for X and for Y axis

jonnieZG:
Text magnification factor can be defined separately for X and for Y axis. It allows you to print "tall" or "wide" text.

Every method that was accepting "textsize" is now overloaded to accept "textsize_x" and "textsize_y".


makermelissa:
Thanks for adding this. Which displays has this been tested on?

jonnieZG:
> Thanks for adding this. Which displays has this been tested on?

I have tested it (and currently using) on a 0.96" I2C SSD1306 B/W OLED.

makermelissa:
Ok, thanks!

makermelissa:
I just merged and then realized that this will break the functions for existing code. I have reverted the merge. Can you submit a new PR but with overloaded functions?

makermelissa:
Nevermind, it's all good. I see that's just what you did.

PR #235, opened 2019-07-03
support AVR8X e.g. ATmega4809

prenticedavid:
rename PORT_t typedef to avoid name conflict
conditional AVR macro for TFT_HARD_SPI

Many Third Party Libraries inherit Adafruit_GFX 
Only Adafruit SPI hardware libraries actually use Adafruit_SPITFT

This PR will allow Adafruit_SPITFT to compile on AVR8X targets e.g. Nano-Every, WiFi-Rev2

I have only tested this on new Arduino Nano-Every with Adafruit_ST7789.
All my other SPI displays are 3.3V.   The Nano-Every is 5V.

David.

### DIFF
--- a/Adafruit_SPITFT.cpp
+++ b/Adafruit_SPITFT.cpp
@@ -35,6 +35,14 @@
 
 #include "Adafruit_SPITFT.h"
 
+#if defined(__AVR__)
+#if defined(__AVR_XMEGA__)  //only tested with __AVR_ATmega4809__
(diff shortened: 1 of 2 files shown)


WillEccles:
For what it's worth, I have tested these changes on a Nano Every (4809-based) with excellent results. While I have not gone back and tested them on my older UNO, I see no reason why these changes would break anything there. Kudos to David for beating me to the punch on the fix ;)

PR #248, opened 2019-10-17
better use nrf spi with dma for drawing pixels with nrf52840

hathach:


ladyada:
verified that DMA is being used for large color sections

PR #253, opened 2019-11-17
Overwriting text

avandalen:
The changes are tested well, see the example.
https://www.avdweb.nl/arduino/hardware-interfacing/ili9341-tft


eudoxos:
Is there something that can be done to merge this change? I find it very valuable (avoids flicker even with custom fonts). The argument at Adafruit's website (that custom fonts have varying glyphs) is moot, it should be left up to the user to user monospace fonts when also using overwriting, or proper space padding.

avandalen:
You are a collaborator now..

PaintYourDragon:
Making a sweep through old PRs. Grateful for the contribution, but we’ve settled on the Official Canonical Way To Handle Non-Blinking Overwrites With Custom Fonts is to use an interim GFXcanvas1 object. Thanks.

PR #254, opened 2019-12-03
Add 16-bit read/write commands for NT35510 display lib

PaintYourDragon:


ladyada:
plz fix errors in ESP?

PR #257, opened 2019-12-19
Change in character size on button label

nestorlevi:
I have been testing the **Adafruit_GFX_Button** function and I noticed that the button label cannot exceed 9 characters. I have changed the library to allow a greater range using the **strcpy()** function instead of **strncpy()**.
Nestor Palomeque.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1575,7 +1575,7 @@ void Adafruit_GFX_Button::initButtonUL(
   _textsize_x   = textsize_x;
   _textsize_y   = textsize_y;
   _gfx          = gfx;
-  strncpy(_label, label, 9);
+  strcpy(_label, label);
 }
 
 /**************************************************************************/


ladyada:
hiya strcpy is not permitted, its unsafe. please find another way to do it!

nestorlevi:
Thanks for your reply.
You can estimate the number of characters the button can contain using its size and text size and then limit the size of the label.
Looking at the drawButton function I see that it positions the cursor using a calculation to center the string, it is **(_w / 2) - (strlen (_label) * 3 * _textsize_x)**.
From there I have obtained an equation that estimates the number of characters: **n = w / (6 * textsize_x).**
I have taken into account the division by 0 taking that value to 1 (being the smallest text size).
I had to use the **strlcpy** function instead of **strncpy** since it didn't generate the end of the string correctly.
Also modify the header tag **char _label [10]** so I have defined it as **char * _label** and dynamically creating the memory with the **new** operator after obtaining the estimated number of characters for that button.
I've tried changing text size, button size and with different label sizes and the text remains inside the button, everything looks great.
What do you think?

Adafruit_GFX.cpp
```diff
@@ -1572,10 +1572,13 @@ void Adafruit_GFX_Button::initButtonUL(
...
   _fillcolor    = fill;
   _textcolor    = textcolor;
-  _textsize_x   = textsize_x;
+  if(textsize_x == 0) textsize_x = 1;
   _textsize_y   = textsize_y;
   _gfx          = gfx;
-  strncpy(_label, label, 9);
+
+  uint8_t len =  _w / (6 * _textsize_x);
+  _label = new char[len];
```
BillyDonahue: `len` is the number of displayed characters that will physically fit inside the button.
But you need space for all the displayed characters PLUS a nul terminator.
    `_label = new char[len + 1];`.

Adafruit_GFX.h
```diff
@@ -286,7 +287,7 @@ class Adafruit_GFX_Button {
   uint8_t       _textsize_x;
   uint8_t       _textsize_y;
   uint16_t      _outlinecolor, _fillcolor, _textcolor;
-  char          _label[10];
+  char          *_label;
```
BillyDonahue: used to be 10 bytes. Now it's 4 bytes for the pointer, plus 4 bytes for the internal allocated array size, plus 4 bytes allocator node overhead plus alignment overhead. So 12 of overhead PLUS the length of the label itself, so a 9-char button goes up from 10 to 28 bytes. Maybe just change the 10 to something bigger if you need something bigger. Maybe make it a static constant to be tweaked. Button labels aren't going to get that big.

Issue #266, opened 2020-01-06
drawChar() assumes 8 bit width and TomThumb font contains wrong glyph width


BlueAndi:
I tried to implement a function to align text with the help of getTextBounds(), using the font TomThumb.

Using only "A" will return 8 pixels as boundary width. Using a "AA" will return 12 pixels and using "AAA" will return 16 pixels. Calculation looks like its using the width for the first character + xAdvance for the following characters. But shouldn't it step forward always with xAdvance?

Another problem is that all glyphs in TomThumb has a width of 8 pixels. But as defined the width shall corresponds to the bitmap dimension, that represents the character.

And another problem is in the drawChar() routine:
https://github.com/adafruit/Adafruit-GFX-Library/blob/c16c8dca43c62ceb7c4fd834dad2799887425f3d/Adafruit_GFX.cpp#L1216-L1231

@PaintYourDragon The drawChar() routine assumes that the width is always 8, because ```bit``` is only cleared in this case. But if the width is lower than 8, ```bit``` is not set to 0 for the next row in y direction.
Fixing this and the fixing the width and xAdvance in TomThumb shows it's working perfect.

BlueAndi:
@PaintYourDragon Argh ... I see the problem. All other fonts than TomThumb are "packed". That means TomThumb must be converted accordingly.

PR #268, opened 2020-01-09
TomThumb font now compressed to fix bounding box text width problem

BlueAndi:
Bitmaps now compressed (like the other fonts), to fix the bounding box problem, because non-compressed the calculated text width were wrong. Additional advantage is that it needs less program space.

#266


Fonts/TomThumb.h
```diff
@@ -43,429 +43,435 @@
 ** https://gist.github.com/skelliam/322d421f028545f16f6d
 ** William Skellenger (williamj@skellenger.net)
 ** Twitter: @skelliam
-**
+** 
+** Jan 09, 2020: Bitmaps now compressed, to fix the bounding box problem,
```
BillyDonahue: Can you attach the code you used to generate the font header?
It's good to be able to generate the header from the upstream font source BDF file.
We previously had a link to the @skelliam bdf2adafruit.py gist.

I would like to suggest that the @skelliam bdf2adafruit.py be modified or rewritten to produce your improved header format, and then checked into the fontconvert/ directory here, so we never lose access to it.

Fonts/TomThumb.h
```diff
@@ -28,13 +28,13 @@
 ** THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 ** (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
 ** THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-**
-** Modifications to Tom Thumb for improved readability are from Robey Pointer,
+**
```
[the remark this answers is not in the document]
BlueAndi: *autsch* I wondered why clang check failed.

Fonts/TomThumb.h
BlueAndi: I changed the bdf2adafruit.py gist a little bit, to generate the compressed set. ;-)
Will commit it to /fontconvert folder.

BlueAndi:
@BillyDonahue The CI checks still fails (clang part), but I don't know why. Can you give me a hint?

BillyDonahue:
@BlueAndi   If you expand the details on the failed "clang" check, it shows a diff. That's showing the difference between what clang-format wants and what you have. You can clean that up and apply it with `patch` directly.

https://github.com/adafruit/Adafruit-GFX-Library/pull/268/checks

...
2020-01-09T18:29:47.8803821Z --- ./Fonts/TomThumb.h	(original)
2020-01-09T18:29:47.8805049Z +++ ./Fonts/TomThumb.h	(reformatted)

Looks like it wants slightly different padding for the end-of-line comments.
And it wants to bin-pack the fields of the TomThumb GFXFont.

    -const GFXfont TomThumb PROGMEM = {
    -    (uint8_t  *)TomThumbBitmaps,
    -    (GFXglyph *)TomThumbGlyphs,
    -    0x20, 0x7E, 6 };
    +const GFXfont TomThumb PROGMEM = {(uint8_t *)TomThumbBitmaps,
    +                                  (GFXglyph *)TomThumbGlyphs, 0x20, 0x7E, 6};

The easiest way to fix this might be to download clang-format 3.8.0 and run it yourself.

BlueAndi:
@BillyDonahue Yeah, I did it. Is it possible that the CI clang output explains it better? At the end it only shows error. Thats it.

PaintYourDragon:
Thank you!

PR #270, opened 2020-01-27
Set member variables that never changes as const

bfiset:
Set member variables that never changes as const.
They are already assigned on object construction.


ladyada:
thanks - please clang-format :)

makermelissa:
With the ST7789 displays, they need to be initialized with 320x240, but the WIDTH and HEIGHT variables are changed in the init function so that tft.width() and tft.height() return the correct values. Changing this to a const would break that.

makermelissa:
Ok, I got a PR out to fix the ST7789 code so that this doesn't break it. :)

PR #271, opened 2020-01-28
getTextBounds method that uses internal cursor values

bfiset:
Simpler method call that uses current cursor values.
Useful when you want to get the size of string you are about to print.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1475,9 +1475,9 @@ void Adafruit_GFX::getTextBounds(const char *str, int16_t x, int16_t y,
     @param    h      The boundary height, set by function
 */
 /**************************************************************************/
-void Adafruit_GFX::getTextBounds(const char *str, int16_t *x1, int16_t *y1, 
+void Adafruit_GFX::getTextBounds(const char *str, int16_t *x1, int16_t *y1,
                                  uint16_t *w, uint16_t *h) {
-    getTextBounds(str, cursor_x, cursor_y, x1, y1, w, h);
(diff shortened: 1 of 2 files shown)


BillyDonahue:
I think we should not add this, just because it's possible to achieve the same functionality by composing existing member functions without too much trouble. A caller could call the public `getCursorX()` `getCursorY()` and provide those to the existing getTextBounds function.
I think it even ends up being clearer at the call site:

`gfx.getTextBounds(string, gfx.getCursorX(), gfx.getCursorY(), &x1, &y1, &w, &h);`

If I were to instead use the proposed new overload,

`gfx.getTextBounds(string, &x1, &y1, &w, &h);`

To understand I'd have to know that this overload uses the cursor position, and it's not obvious to me.
It could just as easily be using &x1,&y1 as in/out parameters or something.

bfiset:
> I think we should not add this, just because it's possible to achieve the same functionality by composing existing member functions without too much trouble. A caller could call the public `getCursorX()` `getCursorY()` and provide those to the existing getTextBounds function.
> I think it even ends up being clearer at the call site:
> 
> `gfx.getTextBounds(string, gfx.getCursorX(), gfx.getCursorY(), &x1, &y1, &w, &h);`
> 
> If I were to instead use the proposed new overload,
> 
> `gfx.getTextBounds(string, &x1, &y1, &w, &h);`
> 
> To understand I'd have to know that this overload uses the cursor position, and it's not obvious to me.
> It could just as easily be using &x1,&y1 as in/out parameters or something.

Yeah you're right, I ended up implementing it in my own derived class anyway.
Closing

PR #272, opened 2020-02-12
Removed duplicate spiWrite to fix bbSPI

makermelissa:
When examining the code and comparing it to the old working code, I noticed there was a duplicate spiWrite. I removed it and it solved the bbSPI issue I was seeing. I also tested on hardware SPI and that worked fine.

Fixes https://github.com/adafruit/Adafruit-SSD1351-library/issues/27.

I tested on SSD1351 using software and hardware SPI with an Adafruit Metro.


makermelissa:
Before merging this in, I'd like to test using a Feather M4 board just to make sure I didn't mess up anything with 16-bit.

ladyada:
don't worry about testing 16-bit - nothing uses it yet

ladyada:
fyi @PaintYourDragon

PR #279, opened 2020-03-31
Separate decls into their own statements. Move all API docs to header.

BillyDonahue:
(Note that there is no API change here.)

Move all the API documentation from the .cpp file to the header. We previously had a mix where some functions were documented in the header, and this was inconsistent. There are very good docs on these functions, but they can't be browsed easily. One has to flip between .h and .cpp files.

Separate all plural declaration statements into singular declaration statements.
This will clang-format better, allows functions to be grouped more easily, and gives us a place to hang documentation for each function.

Some function declaration syntax changes to be consistent with API syntax, where there are multiple ways to express a type, pick one and use it consistently:

    - `const T arr[]` becomes `const T *arr`.

    - `f(void)` becomes `f()`.

Provide the missing definition for `void Adafruit_GFX::write(uint8_t)` for `ARDUINO<100`.
Both writes are implemented as a trivial forward to a new common `_write` helper function.

Fix minor issues with the API Doxygen formatting here and there.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -99,14 +99,6 @@ inline uint8_t *pgm_read_bitmap_ptr(const GFXfont *gfxFont) {
   }
 #endif
 
-/**************************************************************************/
-/*!
-   @brief    Instatiate a GFX context for graphics! Can only be done by a
-   superclass
-   @param    w   Display width, in pixels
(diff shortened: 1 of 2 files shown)


BillyDonahue:
#287 is a more modest form of this.
Maybe it's better to go in stages.

PR #282, opened 2020-04-18
Change "char" to "unsigned char" in charBounds()

rafw87:
I saw that `getTextBounds()` skips characters above 0x80. Problem lies in `charBounds()`, where character is compared with first and end character in font, but `char` is signed and characters above 0x80 are out of range because are negative.

I've changed signature to `unsigned char` and works for me.

C++ is not my primary language, so please let me know if my approach is wrong and should be fixed in different way.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1364,7 +1364,7 @@ void Adafruit_GFX::setFont(const GFXfont *f) {
     @param    maxy  Maximum clipping value for Y
 */
 /**************************************************************************/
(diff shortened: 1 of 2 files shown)


ladyada:
hihi thanks - can you run clang-format or follow the guidelines here to fix formatting pleaes
https://github.com/adafruit/Adafruit-GFX-Library/pull/282/checks?check_run_id=598316258#step:7:7

rafw87:
Done 🙂

ladyada:
thanks - we merged it in manually (many other changes were made :)

PR #286, opened 2020-05-13
please update gfxfont.h to allow use with UTF-8

ZinggJM:
My goal is to add an extension to use (multiple) UTF-8 fonts in a subclass of Adafruit_GFX.
This extension is based on the fork: https://github.com/Bodmer/Adafruit-GFX-Library.
Having this extension in a subclass would allow to use the actual version of Adafruit_GFX.

Without this modification I have to use tricks with namespaces to use fonts generated by Bodmer's fontconverter.


ladyada:
thankx! :)

EliasA97:
Hello, how can I create a font using greek characters?

PR #287, opened 2020-05-13
Break up the compound decls in the `BASIC DRAW API` section

BillyDonahue:
No technical change. Just separating the multi-member decls into single-member decls.

This helps with clang-format, and conforms with C++ convention. It's hard to scroll 52 lines up from `setFont` to see its associated `virtual void`. Also allows functions to be arranged by functional group rather than their type, and gives a better place to hang per-member documentation.

Fix up the `drawPixel` Doxygen comment, which seems to have been ruined by a previous clang-format run or something.


ladyada:
wow thank you! so pretty now :) will merge once passing CI

PR #290, opened 2020-05-17
Update MonoOLED i2c constructor to match var def order

nsasch:
This updates the initialization of variables in the I2C Adafruit_MonoOLED constructor. The call signature and variable definitions are unchanged. This fixes a `reorder` compiler warning due to member initialization not matching member declaration. This becomes a compiler error with the [standard ESP32 platform.txt](https://github.com/espressif/arduino-esp32/blob/49b76649f13acdb8abaf795015d2740bc5f6b340/platform.txt#L21) with warning level "All".

The mock_ili9341 example compiles with no new warnings with `compiler.warning_level=all` in the latest Arduino IDE. This was tested for Adafruit Circuit Playground, ESP32, and Arduino Due.

### DIFF
--- a/Adafruit_MonoOLED.cpp
+++ b/Adafruit_MonoOLED.cpp
@@ -60,8 +60,8 @@
 Adafruit_MonoOLED::Adafruit_MonoOLED(uint16_t w, uint16_t h, TwoWire *twi,
                                      int8_t rst_pin, uint32_t clkDuring,
                                      uint32_t clkAfter)
-    : Adafruit_GFX(w, h), buffer(NULL), dcPin(-1), csPin(-1), rstPin(rst_pin),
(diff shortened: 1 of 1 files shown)


ladyada:
i should figure out how to turn on `warning_level=all` in the arduino cli so CI can catch it, do you know how by chance?

nsasch:
I'm not sure off-hand. It might require a change in [adafruit/ci-arduino](https://github.com/adafruit/ci-arduino). I'll look into it today, and either make another PR or comment here with what I find. I'm sure it's one line of code... just need to find which line.

ladyada:
yeah ideally we update ci-arduino and then all 250+ arduino libraries will get the new check :)

PR #293, opened 2020-05-22
Add yield hook as a weak definition.

### LINKED ISSUE #285: ESP8266 panic due to yield() in writeLine()
Hi,
with commit d81e3351f3c2a8e280dbbd531f8c20d949e719e8 @PaintYourDragon introduced `yield()` for ESP8266 into `writeLine()` and `drawCircle()`.

```
void Adafruit_GFX::writeLine(int16_t x0, int16_t y0, int16_t x1, int16_t y1,
        uint16_t color) {
#if defined(ESP8266)
    yield();
#endif
```

Does anyone know the reason for that?

The problem with this is that calling `yield()` outside loop, e.g. from within code called by a Ticker / ISR will cause the ESP8266 to panic an reset. Thus, some drawing functions cannot be used with Ticker.

Code to reproduce the issue with `yield()`:
```
#include <Arduino.h>
#include <Ticker.h>

Ticker tick;

ICACHE_RAM_ATTR void tock() {
  yield();
}

void setup() {
  tick.attach_ms(100, tock);
}

void loop() {
}
```

BillyDonahue:
We could leave the `yield()` behavior customizable to handle edge cases like #285. Most ESP8266 users need to call `yield()` inside expensive drawing operations to keep the watchdog from firing. However, some sophisticated users who are using a different scheduling regime like Ticker must not call `yield()`, so we provide an opt-out.

I like the weak symbol for this, over other hook techniques, because the linker can remove the replaced symbol, there's no init-order race, and it doesn't bring any of this drama into the header. :)

Fixes #285

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -39,6 +39,22 @@ POSSIBILITY OF SUCH DAMAGE.
 #include <pgmspace.h>
 #endif
 
+/*!
+   Expensive drawing operations normally call yield() to reset the watchdog
+   on the ESP8266 platform. Defined as a weak symbol to allow customization
+   of that behavior in applications with different scheduling.
+   Example, in some "myApp.cpp":
+
+     // (overriding yield behavior app-wide)
+     void Adafruit_GFX_yield() { if (okToYield) yield(); }
+*/
+void Adafruit_GFX_yield() __attribute__((weak));
+void Adafruit_GFX_yield() {
+#if defined(ESP8266)
+  yield();
+#endif
+}
+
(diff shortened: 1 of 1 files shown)


BillyDonahue:
A few more people have chimed in on #285 since this was created, and I think this small fix might serve everybody pretty well. Promoting from draft to real PR in the interest of solving `yield()` issues forever.

vortigont:
Would be nice to have this little patch merged.

kkuez:
Hi @BillyDonahue , has this been merged?
Kind regards and thank you for taking care.

PR #296, opened 2020-05-27
Ability to override drawChar method

bfiset:
Had to use my implementation for drawing custom fonts more efficiently. Need to re-implement  `drawChar` method in my interface driver.

Drawback is increased vtable size I guess? (Is it even included in vtable if there isn't a subclass implementation?)

### DIFF
--- a/Adafruit_GFX.h
+++ b/Adafruit_GFX.h
@@ -62,6 +62,10 @@ public:
                         uint16_t color);
   virtual void drawRect(int16_t x, int16_t y, int16_t w, int16_t h,
                         uint16_t color);
+  virtual void drawChar(int16_t x, int16_t y, unsigned char c, uint16_t color,
+                        uint16_t bg, uint8_t size);
+  virtual void drawChar(int16_t x, int16_t y, unsigned char c, uint16_t color,
+                        uint16_t bg, uint8_t size_x, uint8_t size_y);
 
   // These exist only with Adafruit_GFX (no subclass overrides)
   void drawCircle(int16_t x0, int16_t y0, int16_t r, uint16_t color);
@@ -104,10 +108,6 @@ public:
                      const uint8_t mask[], int16_t w, int16_t h);
   void drawRGBBitmap(int16_t x, int16_t y, uint16_t *bitmap, uint8_t *mask,
                      int16_t w, int16_t h);
-  void drawChar(int16_t x, int16_t y, unsigned char c, uint16_t color,
-                uint16_t bg, uint8_t size);
-  void drawChar(int16_t x, int16_t y, unsigned char c, uint16_t color,
-                uint16_t bg, uint8_t size_x, uint8_t size_y);
   void getTextBounds(const char *string, int16_t x, int16_t y, int16_t *x1,
                      int16_t *y1, uint16_t *w, uint16_t *h);
   void getTextBounds(const __FlashStringHelper *s, int16_t x, int16_t y,


BillyDonahue:
The differences introduced are more significant than the vtable overhead.
Here's a trivial sketch that just makes an oled display and calls .begin() on it. That's all it does.

```
#include <Adafruit_GFX.h>
#include <Adafruit_SSD1306.h>
#include <Arduino.h>
#include <Wire.h>
Adafruit_SSD1306 oled(128, 32, &Wire);
void setup() { oled.begin(); }
void loop() {}
```

Compiled for Arduino Uno and analyzed with avr-objdump and diffed before and after this change:

```
diff --git a/Adafruit_GFX.h b/Adafruit_GFX.h
index e81dd74..895e512 100644
--- a/Adafruit_GFX.h
+++ b/Adafruit_GFX.h
@@ -104,10 +104,10 @@ public:
                      const uint8_t mask[], int16_t w, int16_t h);
   void drawRGBBitmap(int16_t x, int16_t y, uint16_t *bitmap, uint8_t *mask,
                      int16_t w, int16_t h);
-  void drawChar(int16_t x, int16_t y, unsigned char c, uint16_t color,
-                uint16_t bg, uint8_t size);
-  void drawChar(int16_t x, int16_t y, unsigned char c, uint16_t color,
-                uint16_t bg, uint8_t size_x, uint8_t size_y);
+  virtual void drawChar(int16_t x, int16_t y, unsigned char c, uint16_t color,
+                        uint16_t bg, uint8_t size);
+  virtual void drawChar(int16_t x, int16_t y, unsigned char c, uint16_t color,
+                        uint16_t bg, uint8_t size_x, uint8_t size_y);
   void getTextBounds(const char *string, int16_t x, int16_t y, int16_t *x1,
                      int16_t *y1, uint16_t *w, uint16_t *h);
   void getTextBounds(const __FlashStringHelper *s, int16_t x, int16_t y,
```

The results:

```
diff --git a/master.objdump b/virtualDrawChar.objdump
index 52d0cb7..79e6c87 100644
--- a/master.objdump--syms
+++ b/virtualDrawChar.objdump--syms
-???????? g     O .data	0000002c vtable for Adafruit_SSD1306
+???????? g     O .data	00000030 vtable for Adafruit_SSD1306
-???????? g     F .text	000001b0 Adafruit_GFX::write(unsigned char)
+???????? l     F .text	000001c0 Adafruit_GFX::write(unsigned char)
+???????? g     F .text	0000002a Adafruit_GFX::drawChar(int, int, unsigned char, unsigned int, unsigned int, unsigned char)
```

The `vtable for Adafruit_SSD1306` grew by 2 16-bit pointers as expected.

We can also see that `Adafruit_GFX::write` grew by 16 bytes. Its 2 calls to the now-virtual `drawChar` now take slightly more computation (4 more 2-byte instructions each call) to set up.

The most important change however, is that the unused `drawChar` overload can't be omitted. If it's in a `vtable`, there's a reference to its address, and the linker can't omit it from the executable. One `drawChar` is used by the `write` virtual function, and the linker can't get rid of `write`, so it's dragged into even a do-nothing app like my example. But the other `drawChar` overload with just `size` instead of separate `sizeX`, `sizeY` would have been free to be dropped but now it can't. So that's 42 more bytes for that function. It's kind of stupid, but that's how it is.

You have another issue that once it goes virtual, subclasses will depend on its exact signature and you can't tweak it anymore without breaking them. I think the integer types have evolved.

You might be better served by restructuring your app to draw your chars "manually"?

bfiset:
That's interesting and unfortunate.

What's unfortunate is that the call to `drawChar()` is buried  within `print()` and subsequent methods in Adafruit_GFX. In order to benefit from the facilities of Adafruit_GFX when it comes to printing characters, I would need to reimplement most functions that leads up to the call to `drawChar()`.

Here is my implementation of `drawChar()` that reduces the amount of SPI transactions for custom characters but it's mostly specific to any derived Adafruit_SPITFT implementation. 

```
// Uncomment to make use of umm_malloc to dynamically allocate buffer instead of using stack
//#define DRAWCHAR_USE_DYNALLOC
// Larger lowers SPI transaction count for a single character draw
// Must be at least as large as the width of the largest character intended to draw
static uint16_t const DrawCustomCharMaxBufferSize = 256;

void Zettler_Driver::drawChar(int16_t x, int16_t y, unsigned char c, uint16_t color, uint16_t bg, uint8_t size_x, uint8_t size_y)
{


    // Character is assumed previously filtered by write() to eliminate
    // newlines, returns, non-printable characters, etc.  Calling
    // drawChar() directly with 'bad' characters of font may cause mayhem!
    if (!gfxFont)
    {
        Adafruit_ST77xx::drawChar(x, y, c, color, bg, size_x, size_y);
    }
    else
    {
        c -= gfxFont->first;
        GFXglyph *glyph = gfxFont->glyph + c;
        uint8_t *bitmap = gfxFont->bitmap;;

        uint16_t bo = glyph->bitmapOffset;
        uint8_t w = glyph->width, h = glyph->height;
        uint8_t bits = 0, bit = 0;
        /*int8_t xo = pgm_read_byte(&glyph->xOffset),
               yo = pgm_read_byte(&glyph->yOffset);*/
        /*int16_t xo16 = 0, yo16 = 0;
        if (size_x > 1 || size_y > 1) {
          xo16 = xo;
          yo16 = yo;
        }*/

        RUNTIME_ASSERT(DrawCustomCharMaxBufferSize > w);
        uint8_t const maxBufferUse = (DrawCustomCharMaxBufferSize / w) * w;
#ifdef DRAWCHAR_USE_DYNALLOC
        uint16_t * const charBuf = static_cast<uint16_t * const>(umm_malloc(sizeof(uint16_t) * maxBufferUse));
        RUNTIME_ASSERT(charBuf);
        if(charBuf)
        {
#else
        uint16_t charBuf[DrawCustomCharMaxBufferSize];
#endif
            uint16_t index = 0;

            uint32_t i;
            for(i = 0; i < (h * w); ++i)
            {
                if (!(bit++ & 7))
                {
                    bits = bitmap[bo++];
                }

                if (bits & 0x80)
                {
                    if (size_x == 1 && size_y == 1)
                    {
                        charBuf[index] = color;
                    }
                    else
                    {
                        //writeFillRect(x + (xo16 + xx) * size_x, y + (yo16 + yy) * size_y,
                                          //size_x, size_y, color);
                    }
                }
                else
                {
                    charBuf[index] = bg;
                }
                if(++index >= maxBufferUse)
                {
                    uint8_t const windowHeight = (index) / w;
                    uint16_t const windowsYStart = ((i + 1) / w) - windowHeight;
                    startWrite();
                    setAddrWindow(x, y - h + windowsYStart, w, windowHeight);
                    writePixels(charBuf, windowHeight * w);
                    endWrite();
                    index = 0;
                }
                bits <<= 1;
            }

            if(index)
            {
                uint8_t const windowHeight = (index) / w;
                uint16_t const windowsYStart = ((i + 1) / w) - windowHeight;
                startWrite();
                setAddrWindow(x, y - h + windowsYStart, w, windowHeight);
                writePixels(charBuf, windowHeight * w);
                endWrite();
            }
#ifdef DRAWCHAR_USE_DYNALLOC
            umm_free(charBuf);
        }
#endif
    }
}
```

I doubt it'd be a good idea to integrate it directly in Adafruit_GFX. It's not even complete as it doesn't support character scaling, yet. But it does make printing (larger) characters significantly quicker.

ZinggJM:
@bfiset,

write() is virtual, declared in Print.h

Implement your own write(), then you can have your onw drawChar() called.

Issue #273, opened 2020-02-18
compile error with freetype version 2.10.1


athusrv:
trying to compile fontconvert but see this error:
warning: implicit declaration of function 'FT_Property_Set'; did you mean 'FT_Face_Properties'? [-Wimplicit-function-declaration]
   FT_Property_Set(library, "truetype", "interpreter-version",
   ^~~~~~~~~~~~~~~
   FT_Face_Properties

Looks like the version 2.10.1 (probably 2.10.0 too) expect the fontconvert.c to include these headers as laided out [here](https://github.com/servo/libfreetype2/blob/master/freetype2/docs/CHANGES#L752)

#include FT_MODULE_H
#include FT_CFF_DRIVER_H

After including these headers, the fontconverter was compiled successfully.

ladyada:
hihi please submit a PR :)

BillyDonahue:
fixed now. The comment on #297 didn't use the right magic words to autoclose this on merge.

ladyada:
rad :)

PR #297, opened 2020-05-28
fontconvert.c missing a #include

BillyDonahue:
`fontconvert/fontconvert.c` was missing an include of `FT_MODULE_H`
Fixes Issue #273


ladyada:
thanks :)

PR #302, opened 2020-06-09
Add const specifier for strings

kinafu:
Without the `const` specifier methods would be able to modify immutable strings. 
-  it fixes the compiler warning: 
`warning: deprecated conversion from string constant to 'char*' [-Wwrite-strings]`
-  prevents illegal operations on the passed `char*` strings.

I did not test the changes.
I assume that right now the change is just relevant for the compiler, because the code currently does not illegally modify the passed `char *`. (Although in the future it could come in handy to receive a compiler message upon violations.)


BillyDonahue:
String `"literals"` can be converted to `char*`, but callers who are calculating their button labels from something other than a simple `"literal"` currently have to work around the lack of `const` in these parameter types, with unnecessary casts that might hide other problems in their code. So this should help callers write safer code in practice, not just silence a warning.

PR #315, opened 2020-08-19
Fixed fast line drawing in GFXcanvas* classes

michaelkamprath:
The existing `GFXcanvas8` class implements the `writeFastHLine` method from `Adafruit_GFX`, but does so incorrectly. The bug is that it doesn't properly handle the current rotation setting. It does correctly "rotate" the line start point of the line, but doesn't account for the fact that in rotations 1 and 3 a horizontal line should be drawn as a vertical line in the underlying raw buffer layout, and that in rotation 2 the direction of the horizontal line is reversed when drawing.

This pull request corrects the above bug, and then generally adds fast horizontal and vertical line drawing to all of the `GFXcanvas*` classes. Furthermore, the `GFXcanvas` example script is extended to demonstrate the fast line drawing under all rotations for all `GFXcanvas*` classes.

The general design of this change is to actually implement the `drawFastVLine` and `drawFastHLine` methods from `Adafruit_GFX` in the `GFXcanvas*` classes. These methods first determine whether given the current rotation setting as to whether the drawing to underlying buffer is actually a horizontal or vertical line, and then the coordinates in the raw buffer coordinates for that line. Then private `drawFastRawVLine` or `drawFastRawHLine` methods are called in order to lay down the line in the underlying buffer. Note that the public interface provides "draw" methods rather than "write" methods (as was originally implemented in `GFXcanvas8`) because `Adafruit_GFX`'s implementation of the corresponding "write" methods actually calls the here implemented "draw" methods. 

This change not only improves the performance of drawing vertical or horizontal lines, but also improves the performance for drawing filled shapes as the `Adafruit_GFX` implementation of the filled shape drawing uses the fast line methods.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1884,6 +1884,191 @@ void GFXcanvas1::fillScreen(uint16_t color) {
 
 /**************************************************************************/
 /*!
+   @brief    Speed optimized vertical line drawing
+   @param    x   Line horizontal start point
+   @param    y   Line vertical start point
+   @param    h   length of vertical line to be drawn, including first point
+   @param    color   Binary (on or off) color to fill with
+*/
+/**************************************************************************/
+void GFXcanvas1::drawFastVLine(int16_t x, int16_t y, int16_t h,
+                               uint16_t color) {
+  if ((x < 0) || (x >= width()) || (y < 0) || (y >= height())) {
+    return;
+  }
+
+  if (y + h > height()) {
+    h = height() - y;
+  } else if (h < 0) {
+    // convert negative heights to their postive equivalent
+    h *= -1;
+    y -= h - 1;
+    if (y < 0) {
+      h += y;
+      y = 0;
+    }
+  }
+
+  if (getRotation() == 0) {
+    drawFastRawVLine(x, y, h, color);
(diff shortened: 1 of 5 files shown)


ladyada:
good find! @makermelissa can test it out in a week

makermelissa:
Just an update. I tried out the GfxCanvas demo on the PyPortal, but was not getting any output. I then tried the current version and am still not getting output, so test results are inconclusive. Unfortunately this is the only board I currently have access to test, so I'll give it another go in a few days with a different board.

michaelkamprath:
@makermelissa I am not deeply familiar with the PyPortal, but I see it has a beautiful screen attached. Note that the `GFXdemo` code prints to the Serial output, and not any screen. Basically, it's using the `GFXcanvas*` classes to make a kind of ASCII art. That was done so that it can run an any device and still show off the GFX features.

makermelissa:
Hi @michaelkamprath, I realize that. It just wasn't running on the M4 processor both before and after the change for whatever reason. But since this doesn't seem to have made anything worse, I consider that a separate problem. I have more boards available to me now and will try running it on that this morning.

michaelkamprath:
>  It just wasn't running on the M4 processor both before and after the change for whatever reason. 

Thanks for explaining!

PR #323, opened 2020-11-14
Fixes warning: comparison between signed and unsigned integer

PhilippeArdit:


PhilippeArdit:
Someone else did this better than me

Issue #325, opened 2020-11-20
FillTriangle using Canvas does not work correctly.


KurtE:
There is a forum thread up on PJRC forum:
https://forum.pjrc.com/threads/64589-fillTriange-issue-with-screen-or-buffer-size

Where a user had issues with a fillRectangle not working correctly on an ILI9341 display.   
It screws up two different ways, depending on the actual triangle.  The version they posted filled only part of the triangle.  If you change the one value (in the case that I post) the 310, it goes completely outside of the bounds of the triangle.

Note: It works fine if you draw not using the canvas.  I verified that with your library as well as my own ILI9341_t3n. 
Example sketch that shows the issue.
```
#define USE_CANVAS
//#include <ILI9341_t3n.h>
#include <Adafruit_GFX.h>
#include <Adafruit_ILI9341.h>
#include <SPI.h>
const uint16_t  tft_Color_Black        = 0x0000;
const uint16_t  tft_Color_Blue         = 0x001F;
const uint16_t  tft_Color_Red          = 0xF800;

#define TFT_CS 10
#define TFT_DC  9
#define TFT_RST  8

//ILI9341_t3n tft = ILI9341_t3n(TFT_CS, TFT_DC, TFT_RST);
Adafruit_ILI9341 tft = Adafruit_ILI9341(TFT_CS, TFT_DC, TFT_RST);

void setup() {

  delay(250);

  tft.begin();
  tft.setFont();
  tft.fillScreen(tft_Color_Black);
  tft.setRotation(0);

}

int loop_count = 0;
#ifdef USE_CANVAS
GFXcanvas16 *canvas = new GFXcanvas16 (240, 320);  //
#endif

void loop() {
  loop_count++;
#ifdef USE_CANVAS
  if (loop_count & 1) {
    canvas -> drawTriangle(0, 0, 240, 0, 120, 320, ILI9341_RED);  // This works fine
    canvas -> fillTriangle(0, 0, 240, 0, 120, 310, ILI9341_BLUE);  // This is what stops halfway down screen.
  } else { 
    canvas -> drawTriangle(0, 0, 240, 0, 120, 320, ILI9341_BLUE);  // This works fine
    canvas -> fillTriangle(0, 0, 240, 0, 120, 310, ILI9341_RED);  // This is what stops halfway down screen.
  }
  tft.drawRGBBitmap(0 , 0 , canvas -> getBuffer () , 240 , 320);
#else
  if (loop_count & 1) {
[...]

PR #326, opened 2020-11-20
Fix GFXcanvas16::drawFastRawHLine

KurtE:
This is a quick fix to fix functions like:
```
GFXcanvas16 *canvas = new GFXcanvas16 (240, 320)
canvas -> fillTriangle(0, 0, 239, 0, 120, 310, ILI9341_BLUE)
```
Could all on down to drawFastRawHLine and the
computerd buffer_index > 32767 which is the largest that could be held in int16_t.  So switched to uint32_t also converted the computed value to uint32_t as well as to remove compiler warning.

Removed another compiler warning as well.

This addresses the problem in
#325


ladyada:
thanks :)

PR #337, opened 2021-02-12
update drawXBitmap to draw background pixels, matching drawBitmap functionality

zunkworks:
I modified Adafruit_GFX.cpp and Adafruit_GFX.h to match the functionality of drawBitmap so that it can draw background pixels. In my application I was creating an animation using drawXBitmap, because it did not draw background pixels, each frame would end up stacking on top of each other. 

The workarounds I tried were to do a fillScreen with the background color and then redraw all of the other elements, but this created flickering and poor performance. I also tried redrawing the x bitmap twice, once with the foreground color delaying, then drawing with the background color to erase it, this also had poor performance and flickering. 

So I modified the library by copying some of the drawBitmap code and modifying it slightly so that it continues to work for x bitmaps and draws a background color. This was tested with a 64x64 px 45 frame animation on an Feather M4 Express with a 2.4" TFT FeatherWing (product id 3315), and it worked well both for static x bitmap and the animation.

This will break functionality for existing code using the library since it now requires a background color parameter. Maintaining functionality would require adding more complexity to the code (an additional bool flag parameter, and an if statement to either draw background pixels or not). Please let me know if this would be a better way to do it and I'll implement and test that functionality and update the PR accordingly.

Thank you!

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -846,10 +846,11 @@ void Adafruit_GFX::drawBitmap(int16_t x, int16_t y, uint8_t *bitmap, int16_t w,
     @param    w   Width of bitmap in pixels
     @param    h   Height of bitmap in pixels
     @param    color 16-bit 5-6-5 Color to draw pixels with
+    @param    bg 16-bit 5-6-5 Color to draw background with
 */
 /**************************************************************************/
 void Adafruit_GFX::drawXBitmap(int16_t x, int16_t y, const uint8_t bitmap[],
-                               int16_t w, int16_t h, uint16_t color) {
+                               int16_t w, int16_t h, uint16_t color, uint16_t bg) {
 
   int16_t byteWidth = (w + 7) / 8; // Bitmap scanline pad = whole byte
   uint8_t byte = 0;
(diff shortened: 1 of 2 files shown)


zunkworks:
Updated code formatting to pass all checks.

BillyDonahue:
Hi @zunkworks,

Breaking existing code isn't going to fly according to the "PRIME DIRECTIVE" in the README.md. This change would leave existing `drawXBitmap` callers broken, with no workaround to get their existing transparent background behavior.

A better way would be to add a new `drawXBitmap(x, y, bitmap, w, h, color, bg)` overload and leave the existing `drawXBitmap(x, y, bitmap, w, h, color)` function. You don't have to feel guilty about duplicating code. Unreferenced functions are edited out by the linker, so there's no code complexity increase (unless an app makes calls to your new function AND to the old one). This would match the existing practice established by `drawBitmap`.

This would be the change I'm suggesting, applied atop your branch.
https://github.com/BillyDonahue/Adafruit-GFX-Library/commit/2168152c8acec0762211a01040cf1b1cef7297c2

In the short term, to get around the flickering maybe you can try painting to a `GFXCanvas16`, then getting its buffer and copying it image to your display with `drawRGBBitmap`. You're on a Cortex M4 so this should be fast. The `GFXCanvas16` doesn't do any I/O.

zunkworks:
I created a new branch and implemented it as a function overload. I'm closing this PR and opening another PR.

PR #338, opened 2021-02-22
Add function overload for drawXBitmap to draw a background color

zunkworks:
I modified Adafruit_GFX.cpp and Adafruit_GFX.h to match the functionality of drawBitmap so that it can draw background pixels. This was implemented as a function overload to avoid breaking compatibility with existing code.

In my application I was creating an animation using drawXBitmap, because it did not draw background pixels, each frame would end up stacking on top of each other.

The workarounds I tried were to do a fillScreen with the background color and then redraw all of the other elements, but this created flickering and poor performance. I also tried redrawing the x bitmap twice, once with the foreground color delaying, then drawing with the background color to erase it, this also had poor performance and flickering.

So I modified the library by copying some of the drawBitmap code and modifying it slightly so that it continues to work for x bitmaps and draws a background color. This was tested with a 64x64 px 45 frame animation on an Feather M4 Express with a 2.4" TFT FeatherWing (product id 3315), and it worked well both for static x bitmap images and the animation.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -872,6 +872,44 @@ void Adafruit_GFX::drawXBitmap(int16_t x, int16_t y, const uint8_t bitmap[],
 
 /**************************************************************************/
 /*!
+   @brief      Draw PROGMEM-resident XBitMap Files (*.xbm), exported from GIMP.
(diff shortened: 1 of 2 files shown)


zunkworks:
Previous related PR https://github.com/adafruit/Adafruit-GFX-Library/pull/337

PR #341, opened 2021-03-08
add getters for text size (x and y) , color fg and bg and gfx font

mirecta:
If i want develop some ui and draw text to rectangle i need text size text color and selected font. i have two options
inherit library which is not good idea because i want use it as separable graphics driver . or second option write getters for it.


aremmell:
I just finished doing this same thing, only to come here and see none of the PRs are even looked at 👎

PR #345, opened 2021-04-12
Add Arduino RP2040 board package compatible optimised support

Bodmer:
Add basic level of optimisation for RP2040 processors.

The is compatible with the official Arduino RP2040 board support package.


ladyada:
please clang-format and tag when passing

bmuessig:
Any progress on this?

Phoenix1747:
Second that, it would be nice to have optimized RP2040 support. Is @Bodmer still active?

PR #347, opened 2021-04-18
Adafruit_GFX_Button & Adafruit_GFX changes: pressedcolor, radius, custom fonts.

IanBUK:
This pull request adds functionality to Adafruit_GFX_Button: setting a custom backcolour for when the button is pressed; setting the radius of the rounded corners and fixes an issue with using custom fonts on labels, along with allowing more than 9 characters on a label. The Adafruit_GFX class has been extended to include a getFont() function that exposes the current font.

Adafruit_GFX_Button:
radius property to Adafruit_GFX_Button, to allow buttons to have different sized radius corners. 0 or less results in sharp corners
getLabel() function, exposes the current label.
drawButton can now show a different colour background when the button is pressed.
_pressedcolor field added to hold the background colour of a pressed button..
new overrides to initButton to allow setting of pressedColor.
_maxlabellength used to hold the maximum length of the label - here increased to 40 characters. The property is used in the drawButton function, so changes need making purely to the static const int in the .h file.
drawButton now respects the current font, so the label is positioned correctly.
Adafuit_GFX
getFont() function to get the current font. This allows the Button etc to change behaviour based on the font. e.g. in the Button, we can now better calculate the width of the label correctly and therefore position it better.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1838,7 +1838,8 @@ void Adafruit_GFX_Button::drawButton(bool inverted) {
 	_gfx->setCursor(_x1 + (_w / 2) - (strlen(_label) * 3 * _textsize_x),
                   _y1 + (_h / 2) - (4 * _textsize_y));
   }
-  _gfx->setTextColor(text);
+ 
+ _gfx->setTextColor(text);
   _gfx->setTextSize(_textsize_x, _textsize_y);
   _gfx->print(_label);
 }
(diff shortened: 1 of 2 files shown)


IanBUK:
The build error is produced by clang, but it doesn't give any actual error detail.  This code runs locally!  

How do I find out what clang is failing on?

ZinggJM:
finally decided to watch no longer

IanBUK:
> finally decided to watch no longer

Finally...sorted it. I’m new to GitHub, I normally work with the MS stack and so had never heard of clang or doxygen. Apologies if I bored or frustrated you.

PR #348, opened 2021-05-03
Adding new fonts

WOLFIEEEE:
- **I added two other fonts in font folder which help people to use the digital font as using other font in big size destroy the visibility**  

- **There aren't any knows limitation although the font size of normal font and this font will differ** 

**Why we need this font?**
As many of Arduino projects use a low-resolution screen that can't able to show the proper text( talking about larger font size), pixels breaks and looks pretty bad to solve the issue here is a digital font which makes the Arduino screen look cleaner.


BillyDonahue:
There is a copyright issue it's [this font of the same name](https://fonts2u.com/digital-7-mono.font).

WOLFIEEEE:
my bad I thought we can use them in open source things, it would be a huge help for peoples

PR #356, opened 2021-06-03
finish drawPentagram method

zybzzz:
this maybe bad code, ignore me ...


PaintYourDragon:
Best kept in user code, not sufficiently generic for library inclusion. Interesting though, this is the second pentagram-drawing PR, so I’ll keep an eye out if it’s really a thing. Thanks!

PR #359, opened 2021-08-27
fix: pin value can be higher than 127 so type should be uint8_t

fpistm:
Hi,

Issue was found by several STM32 users.
Several STM32 have more than 127 digital I/0's.
Moreover, STM32 uses some pin alias `PYn `(`PA0`, `PC13`,...) to ease usage, alias with analog capabilities start from 0xC0 so
alias was also unusable to define `TFT_CS`, `TFT_RST`, `TFT_DC`.

To go further, ArdunioCore-API defined it as uint8_t by default:
https://github.com/arduino/ArduinoCore-API/blob/173e8eadced2ad32eeb93bcbd5c49f8d6a055ea6/api/Common.h#L91

So I guess it is more compliant to use `uint8_t`.
-1 default values are kept and check is done on `0xFF`

Tested with SPI ST7789 and ST7735 devices.

### DIFF
--- a/Adafruit_GrayOLED.cpp
+++ b/Adafruit_GrayOLED.cpp
@@ -61,7 +61,7 @@
             allocation is performed there!
 */
 Adafruit_GrayOLED::Adafruit_GrayOLED(uint8_t bpp, uint16_t w, uint16_t h,
-                                     TwoWire *twi, int8_t rst_pin,
+                                     TwoWire *twi, uint8_t rst_pin,
                                      uint32_t clkDuring, uint32_t clkAfter)
     : Adafruit_GFX(w, h), i2c_preclk(clkDuring), i2c_postclk(clkAfter),
       buffer(NULL), dcPin(-1), csPin(-1), rstPin(rst_pin), _bpp(bpp) {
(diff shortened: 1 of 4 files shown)


fpistm:
Hi @PaintYourDragon 
Thanks for your feedback. I know a change like this would impact several libraries as several relies on this one.
Anyway, I think this issue should be managed as it is not inline with Arduino pin type.

`GFX_pin_t `is a good idea but in this case why not simply use the same name than Arduino: `pin_size_t ` ?

> Either approach requires eventually updating every SPITFT subclass, I don’t have a painless solution here. A third also-non-optimal solution would be only pins <= 127 are supported (but IIRC there was some chip or family, maybe it was certain STM32s, where pin numbers also incorporated the PORT # and thus are nearly always big 12-bit values).

As stated, the STM32 core use the full range of `uint8_t `to defined alias pin number (PYn) as analog (0xC0 based) so using those alias will not work. Anyway using Dx or x will work until 127. 

About 12 bits, in STM32 core it does not exists. I take care about this to be Arduino compatible.

yipmc (approved):

erlef10:
When will this changed be merged? 

I have just had a longer debug session due to this issue. 

My finding btw is that the problem is solved by using the digital definitions of the pin ea Dx and not the analog definition PAx.

I do not appose the change, but for future reference and other googlers, using the digital pin definition may solve your issue. At least for the STM32F4 generic definitions.

PR #371, opened 2021-11-08
implement Fontconvert in Python

BillyDonahue:
Reimplements the fontconvert.c program in Python, which will make it more accessible to those who don't have a C compiler and freetype library installed. Verifies the fonts regenerated by the python implementation match the originals. There are very small number of single-pixel differences in output glyphs vs the originals, but these are due to FreeType version skew and visually look like improvements anyway.


combs:
two things I noticed, as an excited rando--it requires `freetype-py` installed via pip, and it yacks if your ttf has an uppercase extension because the regex is case-sensitive. 

example failure:

```
combs@pripyat:~/git/BillyDonahue-Adafruit-GFX-Library/fontconvert$ python3 fontconvert.py ~/Dropbox/Projects/Slide\ Viewer\ \(Deprecated\)/Commodore-64-v6.3.TTF 8
Traceback (most recent call last):
  File "/home/combs/Dropbox/Git/BillyDonahue-Adafruit-GFX-Library/fontconvert/fontconvert.py", line 267, in <module>
    main()
  File "/home/combs/Dropbox/Git/BillyDonahue-Adafruit-GFX-Library/fontconvert/fontconvert.py", line 263, in main
    FontConvert(**opts).render()
  File "/home/combs/Dropbox/Git/BillyDonahue-Adafruit-GFX-Library/fontconvert/fontconvert.py", line 174, in render
    name="{}{}pt7b".format(re.match(".*/(\w*).ttf", self.infile)[1], self.size),
TypeError: 'NoneType' object is not subscriptable
```

the regex also requires a forward slash to be included in the path. example that failed:

```
python3 fontconvert.py commodore.ttf 8
```

worked:

```
python3 fontconvert.py ./commodore.ttf 8
```

in general I would humbly suggest something like `os.path.basename` to handle path parsing instead of a regex...

Once I got over these speed bumps, it worked great. 

Not in a position of authority to comment on this PR being merged in to Adafruit-GFX-Library, but I hope you don't mind me saying that I found your Python port very helpful for me.

PR #372, opened 2021-11-12
Bitmap and mask display capability of Adafruit_GFX_Button

63n0m3:
Hello.
In pull request there is an Adafruit_GFX_Button changed both in Adafruit_GFX.cpp and Adafruit_GFX.h to add the new functionality. The class is fully compatible with previous version. All original varieables unchanged. Added functionality allows for displaying button bitmaps and masks. Depending on which initButton function user calls, from this point on drawButton will draw old style or double bitmap(true/false) style or double mask with rgb565 color style buttons. There is also added Change_Position() function for adjustable menus, and Last_Time_Used_millis varieable for alternative press of a button determination.
On top of this there is a rgb24 .bmp to rgb565 array and 8pixel/byte mask array generator tool in the bmp extractor folder.
Example with the video tutorial and explanation is in the examples. Link to video explaining both from using and code perspective is in the example Custom_Button_Example.


63n0m3:
The example is based on MCUFRIEND _kbv

Issue #223, opened 2019-06-08
Background color for custom fonts


AntonioFromBrazil:
I really know the difficult involved to implement background color for custom fonts since they do not have an uniform size. Maybe one intermediate solution would be implemented only for monospaced custom fonts. It would be usefull when a better characters resolution is required. The user would create monospaced fonts bigger then the classic 5x7 built-in font. Does it make sense ?

makermelissa:
Hi, would it work for you to draw a filled rectangle behind the character before writing the character or does it already have an unchangeable background on the character that would overwrite that?

AntonioFromBrazil:
Thanks for YOUR comment makermelissa. The problem is that if you draw a filled rectangle the screen will flicking but if the rectangle is done at GFX library it does not occurs. This can be proved writing at same x,y many different values with different bg colors using the classic 5x7 built-in font...

makermelissa:
Do you want to try making the change and submitting a PR?

pljakobs:
the font rendering code doesn't know about monospace or not, nor does the font format we use. 
Flickering can be fixed by drawing to a canvas first and then drawing the full canvas to screen. Really, that's the best possible way.

AntonioFromBrazil:
Thanks pljakobs... Do you know some where to find examples of using canvas as you suggest ? I am not succeed in my tryies ! Thanks a lot

KurtE:
For what it is worth,  recently a few of us added GFX font support back into some of our more optimized display libraries for the Teensy 3.x and 4.0 boards.  Currently we have it in my ILI9341_t3n library (https://github.com/KurtE/ILI9341_t3n) as well s a ST7735/ST7789 and ILI9488 library.

Secondary note: All three of these libraries support the ability to set a backing frame buffer (canvas) for those Teensy boards who have enough memory...   You can tun on frame buffer, and do a several draws and then say tft.updateScreen();  and it does logically a drawRect of the whole buffer to the screen.  Also have Asynchronous update function that uses DMA...

With this GFX font support, we added in some speedups of transparent text output.  In particular it does quick checks to see if can output 8, 4, 3, 2 and finally 1 horizontal logical pixel using one logical FilllRect, which helps reduce the number of outputs...

But in addition to this, we added Opaque text output, which appears to be working pretty well.   To do this, required the code to do some special stuff including: 
[...]

PR #375, opened 2022-01-11
Add fillTextRect() function for alignment, clip and bg fill.

PetteriAimonen:
Added new method `fillTextRect()`.

It handles the following:

- Aligns text horizontally and vertically: centered / left / right / top / bottom.
- Clips any part of text that would go outside rectangle.
- Fills background with minimal flicker, with correct handling for overlapping parts of custom fonts.

---

Risks / limitations / drawbacks:

- No modification to existing functions.
- Does not currently handle text scaling or wrapping. These would be possible to add in the future as an overload if someone needs them.
- Quite large function in code size (about 1kB on ARM), but if it is not called it gets optimized away.

---

Example test:

    display.setFont(&FreeSans18pt7b);
    display.fillTextRect("Top left\nLong line is longer than long cat\nShort", 10, 10, 190, 140, Adafruit_GFX::TOP_LEFT, 0xFFFF, 0x3333);
    display.fillTextRect("Mid center\nLong line is longer than long cat\nShort", 10, 160, 190, 140, Adafruit_GFX::MID_CENTER, 0xFFFF, 0x3333);       
    display.fillTextRect("Bottom right\nLong line is longer than long cat\nShort", 210, 160, 190, 140, Adafruit_GFX::BOT_RIGHT, 0xFFFF, 0x3333);
    display.setFont(&Z003_MediumItalic20pt7b);
    display.fillTextRect("Overlap\nworks also", 210, 10, 190, 140, Adafruit_GFX::MID_CENTER, 0xFFFF, 0x3333);

gives the following output:

Example screenshot

---

Regarding issue #223: This function handles the problem of overlapping glyph parts by tracking the largest `x` and `y` of glyphs rendered so far. The background is written only in new area. Flicker only occurs in the overlapping parts of glyphs, otherwise each pixel is only written once.


PetteriAimonen:
If there are any concerns or questions about this, feel free to ask me. I think this could be quite useful to get into the code.

PetteriAimonen:
@makermelissa Just a ping, would you be interested in taking a look at this pull request? If I understand correctly, you work at Adafruit and had previously commented on issue #223.

PR #376, opened 2022-01-16
Enable Adafruit-GFX-Library as ESP-IDF component support

solhuebner:
Enables Adafruit-GFX-Library as ESP-IDF component support  (with Arduino as component)


ladyada (approved):
why cant it be installed thru the arduino lib manager? i think folks use it that way now?

solhuebner:
Hi @ladyada. I should have been more verbose ;) With this PR the library can be installed via the Arduino board manager (as usual) but also be built via CMake now. That is especially useful if you want to use the [ESP-IDF](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/get-started/index.html) for the ESP32 (and its variants) for example, as they make use of CMake. That way you have more control but you also need to add all the 3rd party libraries as "components". Probably the most useful to start with is [ESP32-Arduino](https://github.com/espressif/arduino-esp32) but if you also want to use proven and loved libraries (like this one) you need to add the CMake files manually if they do not exist (what took me a bit to understand how it works). Some libraries like [ESP32-HUB75-MatrixPanel-I2S-DMA](https://github.com/mrfaptastic/ESP32-HUB75-MatrixPanel-I2S-DMA) already implemented support, so I hope this one will too :)

ladyada (approved):
OK

ladyada:
im ok with it, but we probably will not remember to maintain it if something changes because we dont use CMake! :)

solhuebner:
Thanks :)

PR #383, opened 2022-03-08
Speed up internal font drawing for opaque background color

bitbank2:
This change adds a few lines of code to the drawChar() function to make it more efficient on TFT displays. For characters with an opaque background color, it's much more efficient to create a single memory window for the whole character and write all the pixels into it than to write small rects for each block of pixels. In an older version of the Adafruit_GFX library that I was working with, this change resulted in a > 9x speedup. In the latest GFX lib, something happened to the writeColor() function that reduced the speedup to less than 2x. Still worth doing, but it would be nice to see what it isn't as fast as it used to be. Here's a photo of my test on an older GFX lib before and after my change.
adafruit_gfx

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -246,6 +246,36 @@ void Adafruit_GFX::writeFillRect(int16_t x, int16_t y, int16_t w, int16_t h,
 
 /**************************************************************************/
 /*!
+   @brief    set the address window (memory area), overwrite in subclasses if needed
+*/
+/**************************************************************************/
+void Adafruit_GFX::setAddrWindow(unsigned short x, unsigned short y, unsigned short w, unsigned short h)
+{
+  (void)x; (void)y; (void)w; (void)h;
+}
+
+/**************************************************************************/
+/*!
+   @brief    write len pixels of the given color, overwrite in subclasses if needed
+*/
+/**************************************************************************/
+void Adafruit_GFX::writeColor(uint16_t color, uint32_t len)
+{
+ (void)color; (void)len;
+}
+
+/**************************************************************************/
+/*!
+   @brief    write a buffer of pixels, overwrite in subclasses if needed
+*/
+void Adafruit_GFX::writePixels(uint16_t *colors, uint32_t len, bool block,
+                   bool bigEndian)
+{
+  (void)colors; (void)len; (void)block; (void)bigEndian;
+}
+
+/**************************************************************************/
+/*!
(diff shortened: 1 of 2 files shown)


bitbank2:
It looked like the original build problem was on AVR targets. I removed all new code for AVR with #ifdef. Now I have no idea why it's failing, the error log doesn't show an actual error.

ladyada:
click on the red X to see error log
https://github.com/adafruit/Adafruit-GFX-Library/runs/5466070355?check_suite_focus=true
then find red X failures and click the expander to see the compiler error
image
we find it faster to test the compilation by changing the target on the home computer IDE than resubmitting PR changes

bitbank2:
I did that and it builds fine on my machine with an AVR target. Any other ideas?

bitbank2:
Something strange is happening. I am using uint16_t, not "unsigned short" and the errors above are for compiling an older version of my checkin.

ladyada:
yeah we were compiling the first version, before the ifdef, since the ifdef should be remvoed and the avr compilation fixed?
the line with 'unsighned short' is here?

https://github.com/adafruit/Adafruit-GFX-Library/pull/383/files#diff-1c8be0e9f27e793b8f53a91d73555c8e01f5bce6ff4ea785beee1fd821f46506R253

that looks like code you're adding in the PR?

bitbank2:
This is quickly becoming a THX1138 situation. I'm just about out of time to work on this. If you would like to increase the speed of text drawing in your Adafruit_GFX library, I can explain what I did. It's really a tiny bit of code.

ladyada:
why don't ya start by reverting the #ifdef part - i think thats making it confusing. ifdefs should not be required :)

ladyada:
close, theres still an ifdef chunk here!
https://github.com/adafruit/Adafruit-GFX-Library/pull/383/files#diff-1c8be0e9f27e793b8f53a91d73555c8e01f5bce6ff4ea785beee1fd821f46506R1179

bitbank2:
Rage level rising; that last #ifdef disables the new code. It should build without errors even without that section.

ladyada:
ok it compiles now, just need to clang-format and any doxygen alerts

bitbank2:
Passed all checks; please have a look.

ladyada:
good work! @makermelissa can review when she's got time :)

makermelissa:
Tested on a Metro M4, Metro ESP32-S2 (with pin number changes), and Metro 328 with the ILI9341 shield. It worked fine on all 3 architectures.

makermelissa (approved):

PR #386, opened 2022-04-26
Add bool dmaBusy()

homeodor:
Hi!

I’ve added a tiny function called dmaBusy(). It does what is says on the label: it lets the user code know if DMA is actually on and doing something. Maybe there has been a way to know that before, then I’m sorry for being silly, but I couldn’t find it.

Imagine a situation where you might be okay with skipping some frames instead of waiting on DMA. Say you have some kind of reasonably time-critical device (I have a MIDI controller) and there is some not-so-time-critical UI on screen: it’s perfectly good to update a gauge on the next app cycle; or imagine there is a screensaver with some background calculations (say, it’s “Game of Life”): you might do calculation, but skip the rendering. That’s exactly how I use this.

Thank you for the awesome work!

### DIFF
--- a/Adafruit_SPITFT.cpp
+++ b/Adafruit_SPITFT.cpp
@@ -1172,6 +1172,19 @@ void Adafruit_SPITFT::dmaWait(void) {
 }
 
 /*!
+    @brief  Check if DMA transfer is active. Always returts false  if DMA
+            is not enabled.
+*/
+bool Adafruit_SPITFT::dmaBusy(void) const
+{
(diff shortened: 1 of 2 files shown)


ladyada:
hiya! thanks so much for submitting a PR! we can review & merge PRs once they have passed continuous integration (CI). that means that code compiles cleanly for all tested platforms, is clang formatted so we maintain the same text formatting for all new code, and is doxygen documented.
if your code isnt passing, check the CI output (click on the red X next to the PR to scroll through the log and find where the error is

homeodor:
Hi @ladyada, 

I think I’ve fixed all the issues, so good to go anytime you’re ready.

Thanks a lot!

PR #387, opened 2022-04-27
Added the ability to define a text window. 

ukmaker:
This PR adds the ability to define a window for text. Subsequent calls to print(), write() etc will cause the text to be printed within the window. Lines which are too long to fit will be clipped, if wrapping is not is enabled, at whole character boundaries.

Added functions allow to set or remove the window.
Additional helper functions have been added to get the current font, the font height and whether wrapping is enabled.
getTextBounds() has been modified to take account of windowing.

These changes simplify text widget code.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -117,6 +117,8 @@ Adafruit_GFX::Adafruit_GFX(int16_t w, int16_t h) : WIDTH(w), HEIGHT(h) {
   wrap = true;
   _cp437 = false;
   gfxFont = NULL;
+  textWindowed = false;
+  textX = textY = textW = textH = 0;
 }
 
 /**************************************************************************/


BillyDonahue:
Hi ukmaker,

This is adding some state to the GFX base class and a fair amount of code that is added in a way that it cannot be optimized away if the feature isn't used. I would guess this adds perhaps few hundred bytes to the library footprint?

The new data members would need to be initialized in the constructor, otherwise they take garbage values.

The text-widget windowing can be accomplished today by writing text onto a `GFXCanvas` and then drawing that `GFXCanvas` as a bitmap over the real `Adafruit_GFX` object. That strategy should be simpler to use than this change, and users who aren't interested in drawing clipped text wouldn't have to spend their progmem on it.

If there's going to be a clipping window API, would it be simpler if it applied to all draw functions rather than just the `write(ch)` function? That would be just like using a scratch Canvas though, so clients could use that instead.

ukmaker:
Hi Billy,

My thinking here was the the average UI on a small screen embedded application is a few well defined  widgets and text labels. As such, the only thing which really needs clipping support is the text rendering, so adding it to everything would mean imposing a larger code burden in both size and overall performance. Having text-clipping code in the font rendering engine means I don't have to put it in my application code and hence probably saves progmem in the long run. Plus, setting aside precious RAM just for text rendering makes me sad :-)

As an aside, is there a reason why glcdfont hasn't been rewritten as a new-style font? That way the old font rendering path could be removed which would save progmem and execution time.

PR #413, opened 2023-02-03
Fix multiple non-user-facing typos and grammar

Semnodime:


Semnodime:
The clang check apparently fails. The previous fix-typo branch caused no issues at all. Any ideas?

PR #434, opened 2023-10-10
Remove volatile from wait loops

caternuson:
Removes the `volatile` qualifier being used in a few wait loops. Doesn't seem necessary. This fixes a compile warn for ESP32. Example warn for ref:
```
  /home/runner/Arduino/libraries/Adafruit_GFX_Library/Adafruit_SPITFT.cpp:2309:39: warning: '++' expression of 'volatile'-qualified type is deprecated [-Wvolatile]
   2309 |   for (volatile uint8_t i = 0; i < 1; i++)
        |                                       ^
```

### DIFF
--- a/Adafruit_SPITFT.cpp
+++ b/Adafruit_SPITFT.cpp
@@ -2305,10 +2305,6 @@ inline void Adafruit_SPITFT::SPI_MOSI_HIGH(void) {
 #endif // end !HAS_PORT_SET_CLR
 #else  // !USE_FAST_PINIO
   digitalWrite(swspi._mosi, HIGH);
-#if defined(ESP32)
-  for (uint8_t i = 0; i < 1; i++)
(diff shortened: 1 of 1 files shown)


caternuson:
Tested with a Feather ESP32 and PID 1480 (older non eyespi version) TFT and library example:
https://github.com/adafruit/Adafruit_ILI9341/blob/master/examples/graphicstest/graphicstest.ino

caternuson:
@ladyada for review

ladyada:
i'd just delete - current version will be optimized out

caternuson:
@ladyada take it you mean just remove the loops? pushed those changes and confirmed still working with same test setup.

ladyada (approved):

ladyada:
yep lets try it!

PR #448, opened 2024-05-23
Fixed Adafruit_GFX font size limitation - the 8-bit signed values prevented large fonts from working

bitbank2:
This is a simple change which allows the fontconverter and Adafruit GFX font logic to work with fonts larger than about 90 pt. The original code used signed 8-bit values for the character offsets and yAdvance. These overflow when working with fonts past about 90 points. There is no downside to this fix; adding a few bytes to the size of the GFXglyph structure won't negatively impact anything.


bitbank2:
No idea what the problem is because it just shows "error 1". Useless.

fontconvert/Makefile
```diff
@@ -1,8 +1,8 @@
 all: fontconvert
 
 CC     = gcc
-CFLAGS = -Wall -I/usr/local/include/freetype2 -I/usr/include/freetype2 -I/usr/include
-LIBS   = -lfreetype
+CFLAGS = -Wall -I/usr/local/include/freetype2 -I/usr/include/freetype2 -I/usr/include -I/opt/homebrew/include/freetype2
+LIBS   = -L/opt/homebrew/lib -lfreetype
```
BillyDonahue, line 5: These paths apply to your machine.

The right way to figure out where freetype2 is, would be to have the Makefile use the output of:
`freetype-config --libs` and   `freetype-config --cflags`.

BillyDonahue:
> No idea what the problem is because it just shows "error 1". Useless.

The details are in the failure report:

https://github.com/adafruit/Adafruit-GFX-Library/actions/runs/9206826415/job/25325537261?pr=448

Your code needs be in a state that running clang-format on it would produce no changes.

PR #449, opened 2024-06-05
Fix broken reset logic if reset pin exceeds 127 due to int8 overflow

nocnokneo:
This is in conjunction with adafruit/Adafruit_SH110x#20


nocnokneo:
> @ladyada Any other changes you'd like to see here? adafruit/Adafruit_SH110x#20 (now merged) depends on this PR.

Ping

jepler (approved):
thanks

PR #454, opened 2024-07-09
Make Adafruit_GFX_Button::drawButton work with custom fonts by using …

gilesp1729:
A simple change to make Adafruit_GFX_Button::drawButton work with custom fonts by using getTextBounds to obtain the text bounds. (it is, afaik, also covered by PR #347 in a slightly different way)


gilesp1729:
Virtualising charBounds is really an unrelated change. I don't know how to (or if I can or should) make it the subject of a separate PR.

Issue #415, opened 2023-02-13
 GFX lib with font, canvas, and drawBitmap() runs slow


norbinz:
I breadboarded an Adafruit 3677 (ItsyBitsy 32u4 5V) with an Adafruit 4383 (240x135 LCD ST7789).
I'm seeing very slow run-time of the GFX library's drawBitmap().
In the Adafruit forum I asked what's the reason for the slowness, and does the library offer a faster alternative?
I shared a demonstration sketch that includes the slow drawBitmap(), and a faster code that I whipped up.
A support (?) fellow suggested that I ask on github.
Here's the forum discussion:
https://forums.adafruit.com/viewtopic.php?p=960409

I understand that the GFX lib has a Prime Directive, but I'm not sure if that applies here.

norbinz:
Oops, broken URL. Trying again:
https://forums.adafruit.com/viewtopic.php?p=960155

jsseab:
I noticed this problem too. I have an application that draws a canvas that is 141 x 19 pixels, and this takes around 183 ms with an ATMega644 running at 10 MHz (it's a custom Arduino-compatible board).

I looked at the implementation and realized one big factor is how each pixel write is done with an address window that is 1x1 pixels. This is obviously not a very good idea. A much better implementation is to setup the whole window first, then just write all pixels in sequence.

I changed the code and got about 23 ms execution time (8 times faster!). I will try and submit a pull request for this, but I am a little unsure of exactly how to implement the change due to the class hierarchy.

jsseab:
Any reason why my PR is not accepted yet? Is there anything else that is needed from my side?

ladyada:
we don't have a full time maintainer for this library

PR #458, opened 2024-08-15
Add new method drawBitmapFast with better performance

jsseab:
Added a new method drawBitmapFast in Adafruit_SPITFT which is esentially a copy of the Adafruit_GFX drawBitmap but with much better performance. This is achieved by setting up the address window for the whole image; not for each pixel.

Performance is about 8 times higher.

The method applies clipping so that only the portion of the image that is within the display bounds is written.

Code has been tested with various X/Y offsets to verify that clipping is properly applied.

Note that I was forced to create a copy because of the stated limitations of Adafruit_GFX. I would rather change the Adafruit_GFX method, but that would require adding some abstract methods (e.g. for setting up the window) which would break backward compatibility.

This change is related to this issue: https://github.com/adafruit/Adafruit-GFX-Library/issues/415

### DIFF
--- a/Adafruit_SPITFT.cpp
+++ b/Adafruit_SPITFT.cpp
@@ -957,8 +957,7 @@ void Adafruit_SPITFT::writePixel(int16_t x, int16_t y, uint16_t color) {
    colors. This is a clone of the Adafruit_GFX method but without a call to 
    setAddrWindow for each pixel. Instead the window is setup once, and then
    all pixels are written. This relies on the address auto-increment of the TFT.
-   Note that the image must fit within the display bounds, otherwise nothing is
-   done.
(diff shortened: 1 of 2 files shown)


jsseab:
> Hello, I am very new to GitHub and don't know how pull requests work. If I am using Adafruit-GFX-Library version 1.12.1, will your improvement be included? It looks like the merge has not happened, but again, I don't know what I'm looking at.

It seems like this PR has not been accepted. Not sure exactly why. Either way, you have several options if you want my changes. You could use your existing Adafruit library code and manually apply the changes from my branch - there's only one added function and 2 files changed. That might be the easiest way if you're not used to Git.

Or you could fetch the Adafruit repo and do a local merge of my branch.

Third option is to simply use my branch instead of the Adafruit repo.

Issue #464, opened 2025-02-17
GFXCanvas: Support double buffering and subclass control of framebuffer allocation


jepler:
Over in [PicoDVI](https://github.com/adafruit/PicoDVI/blob/master/src/PicoDVI.cpp#L142), a terrible hack was deployed to permit double buffering of DVIGFX8. This hack seems to have worked well enough but it'd be nice to be rid of it.

Over in [dvhstx](https://github.com/jepler/Adafruit_dvhstx/blob/main/src/Adafruit_dvhstx.cpp#L17), I've taken the (hopefully temporary) approach of copypasting the frame buffer-manipulating routines from this repo into the WIP code.

PR #465, opened 2025-02-18
GFXcanvas: Allow subclass to take charge of buffer allocations

jepler:
In Adafruit_dvhstx, the low level driver takes charge of the framebuffer allocations. Consequently, any allocation by the GFXcanvas class would be an unused waste of memory.

Introduce a new constructor parameter `bool allocate_buffer=true`. The default preserves the old behavior of allocating the buffer in the constructor and freeing it in the destructor, while passing in `false` skips buffer allocation & freeing.

In this case, the subclass must allocate the buffer before the first drawing operation (such as in its constructor or begin method), and free it in its own destructor or end method.

The subclass can also freely change the buffer pointer at runtime, for instance in a swap method if the underlying framebuffer is double-buffered.

Closes: #464

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1758,12 +1758,21 @@ const uint8_t PROGMEM GFXcanvas1::GFXclrBit[] = {0x7F, 0xBF, 0xDF, 0xEF,
    @brief    Instatiate a GFX 1-bit canvas context for graphics
    @param    w   Display width, in pixels
    @param    h   Display height, in pixels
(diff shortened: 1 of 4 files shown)


jepler:
I may need some help with this as there are clang-format changes in files I didn't touch and that I don't reproduce locally (with debian stable running Debian clang-format version 14.0.6). I'm not sure what version of clang-format is being used by ci...

jepler:
ping @ladyada this PR will be a requirement for the HSTX lib I'm working on

Issue #469, opened 2025-04-09
SPI peripheral mapping not working on some RP2040 bases boards


caternuson:
This forum poster was trying to get a [Feather RP2040 CAN](https://www.adafruit.com/product/5724) working with a [TFT FeatherWing](https://www.adafruit.com/product/3651):
https://forums.adafruit.com/viewtopic.php?t=217465

Dealing with ensuring the various CS pins were being handled properly worked to a certain point, however, it was not possible to get a full working example using both CAN and TFT at the same time. I tried to work up an example and ran into a similar issue. Drilling into things, it seems like an issue with some conditional mapping being done to set up the SPI peripheral being used. There are numerous instances of this code:
https://github.com/adafruit/Adafruit-GFX-Library/blob/87e15509a9e16892e60947bc4231027882edbd34/Adafruit_SPITFT.cpp#L1034-L1035

However, it seems the philhower core maps `spi0` and `spi1` to the  RP2040's SPI0 and SPI1 peripherals. The Feather RP2040 CAN's default SPI pins use the SPI1 peripheral:

![Image](https://github.com/user-attachments/assets/6339da09-40fd-4ad4-ae0e-74850e5ef446)

also:
https://github.com/earlephilhower/arduino-pico/blob/e05dd50d626d2871ab988d8e72aa7c12e18e603c/variants/adafruit_feather_can/pins_arduino.h#L32-L37

So it seems like that tertiary logic ends up mapping to `spi0` instead of `spi1` as needed.

Doing a quick hack and changing all instances of:
```cpp
spi_inst_t *pi_spi = hwspi._spi == &SPI ? spi0 : spi1;
```
to:
```cpp
spi_inst_t *pi_spi = spi1;
```
resolves the issue.

So the problem is known, but at this point unclear what a good code fix would be.

**Example test sketch:**
```cpp
/*
 * TFTFeatherWing + Feather RP2040 CAN Example
 */
#include <Adafruit_GFX.h>
#include <Adafruit_ILI9341.h>
#include <Adafruit_MCP2515.h>

#define MCP_CS PIN_CAN_CS   // CAN
#define STMPE_CS 6          // touch
#define TFT_CS   9          // TFT
#define TFT_DC   10
#define SD_CS    5          // SD
   
#define CAN_BAUDRATE (250000)

// CAN
Adafruit_MCP2515 mcp(MCP_CS);

// TFT
[...]

PR #470, opened 2025-04-09
Fix RP2040 SPI peripheral mapping

caternuson:
Potential fix for #469.

Tested with a Feather RP2040 CAN using the example sketch from the issue thread.

Also tested with a basic Feather RP2040 using the unaltered `graphicstest_featherwing` example from the ILI9341 library:
https://github.com/adafruit/Adafruit_ILI9341/blob/master/examples/graphicstest_featherwing/graphicstest_featherwing.ino

Those two boards (hopefully) cover the two general hardware arrangements in terms of SPI peripheral mapping and usage.


caternuson:
@ladyada for review

PR #477, opened 2025-09-13
Added drawEllipse and fillEllipse functions

### LINKED ISSUE #474: Ellipse Primitive
I'm new, and I think I searched for this and I couldn't find anything related so...
Can an ellipse primitive shape be added?  Something like:
`void fillEllipse(int16_t x0, int16_t y0, int16_t rw, int16_t rh, uint16_t color);`
`void drawEllipse(int16_t x0, int16_t y0, int16_t rw, int16_t rh, uint16_t color);`

DS-Aditya-928:
This change adds both drawEllipse and fillEllipse functions using Bresenham's algorithm to the GFX library. The only modifications are the inclusion of both of these functions in the Adafruit_GFX.cpp file and their definitions in the associated header file.

Closes #474


DS-Aditya-928:
@ladyada for review

ladyada (approved):

ladyada:
thanks!

PR #486, opened 2025-11-01
Create overload of fillCircleHelper to be consistent with drawCircleHelper

BillMerryman:
The method fillCircleHelper is inconsistent with the drawCircleHelper method. The brief for fillCircleHelper suggests it can render quarter circle segments while only capable of rendering half circles, while the brief for drawCircleHelper inaccurately states it only supports two mask bits while actually supporting four, one for each quadrant.

This PR addresses these issues by creating an overload of fillCircleHelper to draw each corner independently. A boolean is added to the signature of the overloaded method (useLegacy) that, when true, uses the existing version of fillCircleHelper. When false, it uses a version of the code that is consistent with drawCircleHelper, at the cost of being slightly less efficient (four calls to writeFastVLine instead of two). This approach avoids breaking existing clients. The briefs for fillCircleHelper and drawCircleHelper have also been corrected to more accurately reflect their functionality.

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -640,7 +640,7 @@ void Adafruit_GFX::fillCircleHelper(int16_t x0, int16_t y0, int16_t r,
     @param  color    16-bit 5-6-5 Color to fill with
 */
 /**************************************************************************/
-void Adafruit_GFX::fillCircleHelper(bool useLegacy, int16_t x0, int16_t y0, 
+void Adafruit_GFX::fillCircleHelper(bool useLegacy, int16_t x0, int16_t y0,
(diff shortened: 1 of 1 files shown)


BillMerryman:
@ladyada I originally just modified fillCircleHelper itself, and updated fillCircle and fillRoundRect to account for the changes. Would you be comfortable with that approach?

ladyada:
question: does this even really matter? what is the impact/can you just change the comments?

BillMerryman:
@ladyada I could take the bool out of the signature and rename the method 'fillQuarterCircleHelper'?

ladyada:
lets just change the documentation - why mess with the code if it isnt needed

PR #489, opened 2025-11-07
Add rotated rectangle draw and fill methods

BillMerryman:
These methods implement the ability to draw and fill rectangles rotated by a given angle in degrees.


BillMerryman:
Wasn't sure where you wanted an example, so I placed one in the 'examples' directory in a directory called 'RotatedRectangleTest'. I only have kind of an oddball display to run this on (it is 128 x 128), and it's using the 'Adafruit_ST7735_and_ST7789_Library', so the example is failing the 'test platforms' check (works on my device...). Any suggestions?

Issue #492, opened 2025-11-27
fillEllipse function causes Adafruit Feather ESP32-S3 TFT to hang


bobmoss:
If you use the fillEllipse function on an Adafruit Feather ESP32-S3 TFT board, the board hangs/crashes completely.

It may happen on other ESP32 based boards but I'm unable to check.

The reason is that at the start of the function startWrite() is called. Then drawFastHLine functions are used to draw the lines.

However, drawFastHLine also calls startWrite(). startWrite calls SPI_BEGIN_TRANSACTION() and this nesting of SPI_BEGIN_TRANSACTION causes a deadlock that hangs the board.

The fix is to replace the drawFastHLine calls with writeFastHLine calls as we are already in a "draw" function.

PR #493, opened 2025-11-28
Modified fillEllipse to use writeFastHLine instead of drawFastHLine

bobmoss:
This is a fix for issue #492

I have changed the fillEllipse function to use writeFastHLine instead of drawFastHline. This avoids a deadlock situation with nested startWrite calls.

This follows the same pattern as fillCircle and fillCircleHelper which use writeFast functions and not drawFast functions.


bobmoss:
@ladyada for review.

Issue #490, opened 2025-11-16
GFXfont elements first and last are of type uint16_t, but accessed with pgm_read_byte


bassklampfe:
This will cause issues, when compiling on big-endian architecture.

Should use `pgm_read_word(…)` instead of `pgm_read_byte(…)` and functions like `drawChar(…)` should handle argument `c` and related variables as `uint16_t` and not `unsigned char`.

Became aware of this, when preparing a minimal patch to support UTF8 without big changes to Adafruit-GFX-Library.
My Diff has about 89 lines, where about half of the lines is above issue.

PR #494, opened 2025-12-03
Refactored code to get glyph from a codepoint

bassklampfe:
This fixes https://github.com/adafruit/Adafruit-GFX-Library/issues/490
It also is a preparation for UTF8 support in my fork, but this fix itself is general

- **Describe the scope of your change--i.e. what the change does and what parts
  of the code were modified.**  
  Refactored common code to get GFXglyph from GFXfont. It fixes the issue accessing uint16_t values in GFXfont using (wrong) pgm_read_byte
  By the way it also fixes drawChar, which did not check for limits at all

- **Describe any known limitations with your change.** 
  on atmega platform this patch increases code size by 54 bytes.

- **Please run any tests or examples that can exercise your modified code.** 
 Tested with example from 'Adafruit ST7735 and ST7789 Library' and also with my own code (to be published somewhere in the future)

### DIFF
--- a/Adafruit_GFX.cpp
+++ b/Adafruit_GFX.cpp
@@ -1468,7 +1468,7 @@ void Adafruit_GFX::setFont(const GFXfont *f) {
 /**************************************************************************/
 /*!
     @brief  helper to lookup glyph from gfxFont
-    @param  code   The codepoint (ASCII or unicode) character in question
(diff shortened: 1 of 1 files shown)


bassklampfe:
I did it wrong (old subversion user still learning git). trying to abort this pull request and do it right

PR #503, opened 2026-02-14
Add 8-bit font generation support

NellowTCS:
Hi!
I added optional 8 bit font support to fontconvert because I needed the full 0 to 255 glyph range for a project, and the tool already had everything needed for it. It just wasn't exposed yet.

## What this change does
- Adds support for generating full 8-bit fonts when the user passes first=0 last=255
- Keeps the default 7-bit ASCII behavior the same
- Updates the filename suffix logic so the output is labeled correctly as 7b or 8b
- Updates the docs with the new usage

## What I changed
- Only the character range handling and the documentation.

## Limitations
If the source TTF does not include certain codepoints, FreeType still outputs empty glyphs. This is the same behavior as before.


NellowTCS:
Bump

PR #511, opened 2026-05-13
Overloaded fillCircleHelper to allow for the drawing of filled quarter circles

BillMerryman:
The existing fillCircleHelper implementation does not allow for drawing filled quarter circles, only half circles. It appears that this is the result of fillRoundRect being dependent on this behavior, and fillCircleHelper was written to support fillRoundRect, thus binding their implementations together. This PR creates an overload of fillCircleHelper that removes the delta argument and accompanying implementation such that drawing quarter circles is now possible while avoiding breaking existing clients.


BillMerryman:
@ladyada for review.

PR #517, opened 2026-08-10
Add fillQuarterCircle

BillMerryman:
This method allows for the drawing of filled quarter circles. The repository currently only supports drawing filled half circles vertically.


BillMerryman:
@ladyada for review.

